PR #102, opened 2017-03-21
UI for importing CSV files

jonallured:
This PR adds a basic UI around importing CSV files. You can pick the source you want to import against and then paste in a CSV url from S3 to kick off an Import job. On the show screen, you have to refresh to see progress, but it will tell you when it's done (in-progress vs finished).

What I've done here is just happy path - I haven't done anything about ensuring that errors are handled. My thought was that I'd come at that in a different PR, but maybe that should be here? Not sure what's typical at Artsy for features like this.

Note: this work was done on top of #101 so that it didn't look completely dumb. But it still does look pretty dumb. 😝

### DIFF
--- a/app/controllers/imports_controller.rb
+++ b/app/controllers/imports_controller.rb
@@ -1,0 +1,21 @@
+class ImportsController < ApplicationController
+  expose(:import)
+  expose(:import_result) { ImportResult.new(import) }
+
+  def create
+    if import.save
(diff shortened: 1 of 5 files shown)


starsirius:
Niiiice! Would love to see some screenshots or gifs! 😛 

giphy 1

jonallured:
bearden-walkthrough

@starsirius here ya go! 😄

madeleineb:
It's a real thing!
200w_d

PR #119, opened 2017-03-23
Dynamic import show page

jonallured:
This PR adds ActionCable to the app and uses it to update the counts on an Import show page. It's actually pretty fun to see the numbers zoom along as workers process the import! 😄


gnilekaw:
🚠 awesome

PR #123, opened 2017-03-23
UI improvements

jonallured:
Would love feedback on that simple validation - cool or lame?


jonallured:
Ok, with 5ee7feb I added those required attrs.

PR #126, opened 2017-03-27
Add annotate gem to annotate db schema

starsirius:
Adding one of my favorite gems, [annotate](https://github.com/ctran/annotate_models), to annotate db schema in various places, e.g. models and fabricators, and making them a bit self-explanatory. By default, annotate will be automatically run whenever we run `rake db:migrate` (only in development mode).

Also, updating some gems and fixing Rubocop warnings.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -41,17 +41,20 @@ GEM
       tzinfo (~> 1.1)
     addressable (2.5.0)
       public_suffix (~> 2.0, >= 2.0.2)
+    annotate (2.7.1)
+      activerecord (>= 3.2, < 6.0)
+      rake (>= 10.4, < 12.0)
     arel (7.1.4)
     ast (2.3.0)
     autoprefixer-rails (6.7.7.1)
       execjs
-    aws-sdk (2.8.8)
-      aws-sdk-resources (= 2.8.8)
-    aws-sdk-core (2.8.8)
+    aws-sdk (2.8.11)
+      aws-sdk-resources (= 2.8.11)
+    aws-sdk-core (2.8.11)
       aws-sigv4 (~> 1.0)
       jmespath (~> 1.0)
-    aws-sdk-resources (2.8.8)
-      aws-sdk-core (= 2.8.8)
+    aws-sdk-resources (2.8.11)
+      aws-sdk-core (= 2.8.11)
     aws-sigv4 (1.0.0)
     bootstrap (4.0.0.alpha6)
       autoprefixer-rails (>= 6.0.3)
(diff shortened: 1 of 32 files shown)


Gemfile.lock
```diff
@@ -178,7 +181,7 @@ GEM
       rake (>= 0.8.7)
       thor (>= 0.18.1, < 2.0)
     rainbow (2.2.1)
-    rake (12.0.0)
+    rake (11.3.0)
```
starsirius: Had to update rake to avoid an error:

```
NameError: uninitialized constant Gem::LOADED_SPECS_MUTEX
  /Users/starsirius/.rvm/gems/ruby-2.4.0@global/gems/bundler-1.14.6/lib/bundler/rubygems_integration.rb:426:in `block in replace_bin_path'
  /Users/starsirius/.rvm/rubies/ruby-2.4.0/bin/rake:22:in `<top (required)>'
```

app/models/raw_input_changes.rb
```diff
@@ -10,7 +10,7 @@ def initialize(raw_input)
     @attrs = raw_input.transform
     @organization = nil
     @relations = {}
-    @relations_to_build = [:email, :location, :organization_name, :phone_number]
+    @relations_to_build = %i(email location organization_name phone_number)
```
starsirius: I am not super sure about this rule (as I am very used to the `:symbol` syntax), but this does shorten the length.

jonallured:
Oh man, sorry @starsirius, I am not a fan of Annotate. :(

I find the comments to be quite noisy, but more than that, I think anything that discourages a developer from cracking open a database console should be avoided. I understand that these comments can be convenient but there's a better way to know what columns back an ActiveRecord model:

```
$ rails db
bearden_development=# \d sources
                                           Table "public.sources"
         Column         |            Type             |                      Modifiers
------------------------+-----------------------------+------------------------------------------------------
 id                     | integer                     | not null default nextval('sources_id_seq'::regclass)
 name                   | character varying           |
 created_at             | timestamp without time zone | not null
 updated_at             | timestamp without time zone | not null
 email_rank             | integer                     |
 location_rank          | integer                     |
 organization_name_rank | integer                     |
 phone_number_rank      | integer                     |
 website_rank           | integer                     |
Indexes:
    "sources_pkey" PRIMARY KEY, btree (id)
```

These details are as complete as possible, never out of sync and now that you have opened a console, you can interact with the data should you need to. The database is the source of truth for this information and it's easy to consult it!

But I would happily merge the other parts of this PR - the upgrade to rake and RuboCop. ❤️

starsirius:
😢 

Totally agreed on encouraging developers to open a db console as much as possible. In the end I see the gem as an easy/automatic way to provide db-level "documentation" for the system, and it could be useful for my laziness ( 😛 ) and could also be helpful for communication purposes with non core developers or others. Looks like we also decided to not include the schema in version control. I would still suggest we move, in some way, towards a bit more documentation in general for this type of data-focused project.

But for this one it's your call and feel free to close, and I will PR again for the Rubo-👮  stuff! Thanks!!

jonallured:
Yeah, you're right @starsirius, this project is light on documentation - the ERD wasn't even up to date! I have addressed this with #141, but I'm sure there's more we can do. For now, I'd like to wait on the documentation front until I feel like things are a little more settled. Going to close this one for now. ❤️

PR #127, opened 2017-03-28
UI for sources

jonallured:
One more piece of the UI puzzle - now Importers can view and create Sources. That last point gets a little interesting because when you create a Source that shifts the ranks of the other Sources we have to adjust things. My answer to this tricky problem is the humble `SourceResolver`. It's got a pretty simple interface - it takes the newly built, but not saved, Source and figures out which other Sources need adjustment and to which ranks. Neato!

The other thing I'd point out is the refactor of `rank_for` on Source - now instead of taking a `Rankable`, it takes a symbol, which is great because it let me remove some `public_send` that had me feeling bad about myself.

Note: I put this PR on top of #123 so the diff here looks bigger than it actually is. ❤️

### DIFF
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -10,12 +10,14 @@ class Source < ApplicationRecord
   validates :website_rank, presence: true, uniqueness: true
 
   def rank_for(type)
-    {
+    types = {
       email_rank: email_rank,
       location_rank: location_rank,
       organization_name_rank: organization_name_rank,
       phone_number_rank: phone_number_rank,
       website_rank: website_rank
-    }[type] || raise(UnknownRankType)
+    }
+
+    types[type] || raise(UnknownRankType)
   end
 end
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -1,5 +1,5 @@
 class Source < ApplicationRecord
-  class UnknownRankable < StandardError; end
+  class UnknownRankType < StandardError; end
   has_many :imports
 
   validates :name, presence: true, uniqueness: true
@@ -9,14 +9,13 @@ class Source < ApplicationRecord
   validates :phone_number_rank, presence: true, uniqueness: true
   validates :website_rank, presence: true, uniqueness: true
 
-  def rank_for(rankable)
-    case rankable
-    when Email then email_rank
-    when Location then location_rank
-    when OrganizationName then organization_name_rank
-    when PhoneNumber then phone_number_rank
-    when Website then website_rank
-    else raise UnknownRankable
-    end
+  def rank_for(type)
+    {
+      email_rank: email_rank,
+      location_rank: location_rank,
+      organization_name_rank: organization_name_rank,
+      phone_number_rank: phone_number_rank,
+      website_rank: website_rank
+    }[type] || raise(UnknownRankType)
   end
 end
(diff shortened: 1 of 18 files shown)


spec/rails_helper.rb
```diff
@@ -21,6 +21,12 @@
     PaperTrail.whodunnit = 'Test User'
   end
 
+  config.before(:each, type: :feature) do
```
gnilekaw: 👍

app/models/source_resolver.rb
```diff
@@ -0,0 +1,34 @@
...
+    new(new_source).resolve
+  end
+
+  def initialize(new_source)
+    @new_source = new_source
+  end
+
+  def resolve
+    return if @new_source.save
+    adjust_ranks
```
gnilekaw: Should lines 12 and 13 be wrapped in a transaction so that if something fails everything rolls back?

app/models/source_resolver.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def adjust_ranks
+    taken_rank_types.each do |rank_type|
+      sources_to_adjust(rank_type).each do |source|
+        source.increment!(rank_type)
+      end
+    end
+  end
+
+  def taken_rank_types
```
gnilekaw: This is cool how you save, read the errors, and increment. The word `taken` took me a minute to understand. Although I'm not sure there's a better word to describe what's going on. "Occupied" is a synonym but doesn't make sense. 

They're basically conflicts that are being resolved. Maybe there's a better metaphor there... Something like the following:

```ruby
def resolve_ranks
  conflicted_rank_types.each ...
```

app/models/source.rb
```diff
@@ -9,14 +9,13 @@ class UnknownRankable < StandardError; end
...
-    else raise UnknownRankable
-    end
+  def rank_for(type)
+    {
+      email_rank: email_rank,
+      location_rank: location_rank,
+      organization_name_rank: organization_name_rank,
+      phone_number_rank: phone_number_rank,
+      website_rank: website_rank
+    }[type] || raise(UnknownRankType)
```
gnilekaw: I like `}[type]`, but it feels a little cryptic. Perhaps consider pulling out the hash into a local.

jonallured: Good call - fixed with 0eaa555.

app/models/source_resolver.rb
jonallured: You have the best words - fixed with 96f6105.

app/models/source_resolver.rb
jonallured: This is interesting. You're probably right that a transaction does improve the safety of this code, but I'm having a hard time writing a failing test that proves the need. Open to ideas. Or I could just add the transaction, but it feels like anticipating a problem rather than responding to one, ya know?

jonallured: Ok, I did think of one situation, but it's pretty edgy - what if two Importers were adding Sources simultaneously? A transaction around this bit of code would prevent them from colliding. Probably.

gnilekaw: I don't know what could make this fail. I imagined having to clean it up though when it did. I'm good if you're good.

PR #128, opened 2017-03-28
Import states

jonallured:
Prior to this PR an Import only had an implicit state - something that was reflected by the way the `ImportResult` worked. We'd simply count how many of an Import's `RawInput` records had a nil state and when that number was zero then we knew that the Import was finished.

That worked, but isn't going to cut it moving forward.

Instead, what @gnilekaw and I talked about late last week was adding a state column to the Import and being more declarative about the various states it can be in. To accomplish this, I reviewed the libraries one could use and found something that I really like: [micromachine](https://github.com/soveran/micromachine). It's a tiny little library compared to the competition, which is something that I really like about it - you can easily read the whole thing because it's just [one class](https://github.com/soveran/micromachine/blob/master/lib/micromachine.rb).

So I played around with it and it turns out I think it's a great fit for the states of an Import. I subclassed with `ImportMicroMachine` to encapsulate the configuration so the impact on the Import class is about as trim as I could make it. But make no mistake, this PR *does* add responsibility to Import. It is now the owner of the knowledge about how states and jobs relate to one another. Rather than that knowledge being spread around the system, I've centralized it on the model. You mad?

While this is a fairly large PR, it's actually more of a refactoring than anything else - all I've really done is shuffle around some things. But, what this PR does is set us up with a better foundation on which to build. From here we can introduce the concept of finalizing an Import where we export errors to S3 and ping the team on Slack.

Introducing a pre-flight step is as easy as registering new states in the machine and then writing the job to do it.

And the real punchline will be when we're ready to add the state for syncing with RedShift - we'll be able to tell if an Import's data is in sync because we'll have done the work to ensure our states make sense.

Final note: this work is on top of #127 which itself is on top of #123 - I just couldn't get this done without stacking things like this. I'll happily rebase as things get merged!!


gnilekaw:
This is awesome. Makes sense to centralize everything in the model. Api reads nice also.

PR #129, opened 2017-03-28
[WIP] Add sync feature

gnilekaw:
This PR adds a feature to sync Bearden to Redshift. The sync will trigger after an import has been completed. The process, from the top:

- Upload an import and save raw inputs
- Transform raw inputs into real records
- Write entire database to CSV (yikes) and save to S3
- Copy data from S3 to Redshift

Perhaps you're wondering why not just skip S3 and insert rows directly to Redshift? Well, for large files the [`COPY` command is more efficient](http://docs.aws.amazon.com/redshift/latest/dg/c_best-practices-use-copy.html). And this file will eventually get fairly large. 

**TODO:**
- tests
- update env.example


jonallured:
This is looking 💪 great!

app/models/sync.rb
```diff
@@ -0,0 +1,68 @@
...
+                FROM '#{source}' \
+                WITH CREDENTIALS '#{Redshift.s3_auth}' \
+                DELIMITER ',' \
+                REGION '#{ENV['AWS_REGION']}' \
+                CSV IGNOREHEADER 1 EMPTYASNULL")
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/MethodLength
+  def export_csv
```
jonallured: FYI, I'm working on extracting this functionality into an object we can reuse. I'll link it up when it done so that you can use it here. 🎩

PR #130, opened 2017-03-28
Add job to finalize import

jonallured:
This PR adds a job to finalize imports which means pinging the Slack channel and exporting any errors to S3. The class responsible for this work is the newly minted `FinalizeImportJob`. It's slightly more wordy than I usually like, but there's a decent amount of stuff that goes into closing out an Import. Open to ideas here.

One way I was able to trim the fat on that job was by extracting a few classes. The first one that popped out was `ImportErrors`. The idea here is that I wanted something to encapsulate the logic around converting a set of `RawInput` records into something that could be easily written to a CSV on S3. There's some gnarly parsing going on in `errors_from_details`, but I couldn't think of a better way to handle it! 🤢 

The `SlackBot` class wraps up the pinging of Slack and gets some of those juicy details out of the job.

Finally, the `S3CsvExport` class is something that's been a long time coming. I mentioned this in #129 because I think it'll fit right in on that PR. The idea is to abstract away the work of taking a nested array, a filename and a header row and converting that to CSV and writing to S3.

While I was at it, I finally did my audit of ENV vs the Rails secrets file so that cleanup is also included in the PR. If you'd prefer for that to come in a separate one, please let me know!

Looking for feedback on 0cf5950 - please tell me there's a better way! /cc @yuki24 

Once this lands on staging, then I think it'll be time for @madeleineb to do some testing there - we won't have full exporting finalized until #129 lands, but I've changed this stuff a lot and would benefit from some smoke testing.

Note: this is on top of #128 and thus looks like more than it really is.


app/jobs/finalize_import_job.rb
```diff
@@ -0,0 +1,37 @@
+class FinalizeImportJob < ActiveJob::Base
```
yuki24: This should inferit from [`ApplicationJob`](https://github.com/artsy/bearden/blob/master/app/jobs/application_job.rb).

app/jobs/finalize_import_job.rb
```diff
@@ -0,0 +1,37 @@
...
+    channel = "import_#{@import.id}"
+    ActionCable.server.broadcast channel, import_result
+  end
+
+  def export_errors
+    ImportErrors.export(@import)
+  end
+
+  def post_to_slack
+    url = Rails.application.routes.url_helpers.import_url(@import)
```
yuki24: I don't think this is bad, but at the same time I'm a bit concerned about globally changing the `host` in `config/environments/**.rb`. One thing you could do is to explicitly specify the `host`:

```ruby
Rails.application.routes.url_helpers.import_url(@import, host: 'http://www.bestdomainever.com')
```

If this is a common pattern in Job classes, we could have the parent `ApplicationJob` class include the entire `Rails.application.routes.url_helpers` module and set a `host`:

```ruby
class ApplicationJob < ActiveJob::Base
  include Rails.application.routes.url_helpers
  default_url_options[:host] = 'http://www.bestdomainever.com' # or use ENV['DEFAULT_HOST']
end
```

So in the `post_to_slack` method you could simply do:

```ruby
SlackBot.post(":tada: #{import_result.name} finished: #{import_url(@import)}")
```

Either way, `ENV['DEFAULT_HOST']` still needs to be set.

app/models/import_errors.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def headers
+    raw_inputs_with_errors.first.data.keys + %w(exception error_details)
+  end
+
+  def errors_from_details(details)
+    details.map do |field, detail|
+      errors = detail['content'].map { |obj| obj['error'] }
+      errors.map { |error| [field, error].join(' ') }.flatten
+    end.flatten.join(', ')
```
yuki24: You can simplify this using the [`Enumerable#pluck`](http://api.rubyonrails.org/classes/Enumerable.html#method-i-pluck) and [`Enumerable#flat_map`](https://ruby-doc.org/core-2.4.1/Enumerable.html#method-i-flat_map) methods:

```ruby
details.flat_map do |field, detail|
  errors  = detail['content'].pluck('error')
  errors.flat_map { |error| [field, error].join(' ') }
end.join(', ')
```

Or even:

```ruby
details.flat_map do |field, detail|
  detail['content'].pluck('error').flat_map { |error| [field, error].join(' ') }
end.join(', ')
```

app/jobs/finalize_import_job.rb
jonallured: Ugh, I always make this mistake as the Vim plugin creates the file and inherits from `ActiveJob::Base` by default. Thanks for catching this! ❤️

app/jobs/finalize_import_job.rb
jonallured: Can you explain why changing the host in `config/environments/**.rb` makes you concerned?

app/models/import_errors.rb
jonallured: Awesome refactor, thanks!!!

app/models/s3_csv_export.rb
```diff
@@ -0,0 +1,33 @@
+class S3CsvExport
+  def self.create(rows, filename, headers)
```
gnilekaw: What do you think about doing `def self.create(rows:, filename:, headers:)`? I know it makes the implementation more verbose, but three's a crowd. Perhaps do this when you add the `acl` in the next PR.

app/jobs/finalize_import_job.rb
```diff
@@ -0,0 +1,37 @@
...
+    message = ":tada: #{import_result.name} finished: #{url}"
```
gnilekaw: I can imagine a future improvement that wraps the result name in a url tag, so we don't display the actual url in slack.

app/models/s3_csv_export.rb
```diff
@@ -18,7 +18,7 @@ def create
   private
 
   def s3_bucket
-    Aws::S3::Resource.new.bucket ENV['AWS_BUCKET']
+    Aws::S3::Resource.new.bucket Rails.application.secrets.aws_bucket
```
gnilekaw: I've been using the "Artsy data" bucket in `.env`, but now that I think about it we should make a practice to use the Bearden staging and production buckets. (Note to self.)

app/jobs/finalize_import_job.rb
yuki24: Sorry for the delay @jonallured. Theoretically, there should only be a single host per app, so it shouldn't be a huge problem even if we change the global config(which is why I didn't think this is that bad). The only thing that sticks in my mind is that AcitonMailer has its own `default_url_options` and there might be a reason why this interface is designed this way. What I could imagine is that having two separate configs would make a clear distinction between the `host` set by the request and ActionMailer's `host`, so you could safely use a custom host name in ActionMailer without accidentally overriding the host set by the request.

But again, it should just work as long as we use a single, consistent host name. You could refactor when you need to use an url helper in another jobs class.

PR #132, opened 2017-03-29
Only create organizations when website is provided

### LINKED ISSUE #99: Orgs with nil websites are imported
A RawInput that includes the following data will create a new Organization. We should only allow orgs to be created when they have a `website`.

```ruby
data:
  {"website"=>nil,
   "location"=>nil,
   "latitude"=>nil,
   "longitude"=>nil,
   "email"=>nil,
   "phone_number"=>"8 (495) 730-44-21",
   "organization_name"=>nil,
   "burden_id"=>"93337"
}
```

jonallured:
I decided to add an exception for this case - that felt the most appropriate. I also opted to leave the `error_details` nil because I want to reserve that field for actual ActiveRecord validation errors. This might mean that an Importer could have to decipher exceptions, but maybe that's ok?

I sorta envision a list of exceptions and some explanation on the wiki so that we can help them out that way. Open to thoughts here!

closes #99

### DIFF
--- a/app/models/raw_input_changes.rb
+++ b/app/models/raw_input_changes.rb
@@ -1,5 +1,6 @@
 class RawInputChanges
   class InvalidData < StandardError; end
+  class NoWebsiteBuilt < StandardError; end
 
   def self.apply(raw_input)
     new(raw_input).apply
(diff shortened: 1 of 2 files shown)


gnilekaw:
👍

PR #143, opened 2017-03-31
Only sync when there is something to do

jonallured:
Prior to this commit we'd be going through all the motions of syncing with Redshift even if there was nothing to do! Good catch @gnilekaw.

Note that while I considered reducing the duplication between `nothing_to_sync?` and `update_finished_imports`, I think it's fine. Check this out:

```ruby
> Import.where(state: ImportMicroMachine::FINISHED).any?
   (0.3ms)  SELECT COUNT(*) FROM "imports" WHERE "imports"."state" = $1  [["state", "finished"]]
=> false
```

That count statement is going to be really fast and if you extracted the `Import.where(state: ImportMicroMachine::FINISHED)` part, then you'd lose ActiveRecord's ability to optimize.

### DIFF
--- a/app/jobs/data_warehouse_sync_job.rb
+++ b/app/jobs/data_warehouse_sync_job.rb
@@ -1,5 +1,7 @@
 class DataWarehouseSyncJob < ApplicationJob
   def perform
+    return if nothing_to_sync?
+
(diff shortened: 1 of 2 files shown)


gnilekaw:
🌮

Issue #136, opened 2017-03-30
Use table with states for Import list page


jonallured:
The import list page is currently a ul, but a table would be more appropriate. That way you could see which imports have finished all in one place. This could be hooked up to a websocket so that it updates in real-time, but even without that it would be useful.

Issue #134, opened 2017-03-30
Switch to navbar for top-level navigation


jonallured:
See this page in the docs:

https://v4-alpha.getbootstrap.com/components/navbar/

This change fixes a couple things:

- the Bearden text is the same size as the main h1 on the page
- the selected state of the nav is the same style as a button, but is not clickable

PR #145, opened 2017-04-03
Misc UI tweaks

gnilekaw:
Add a table to display imports on the imports page more clearly.

Closes https://github.com/artsy/bearden/issues/136
Closes https://github.com/artsy/bearden/issues/134

## Imports: Before

screen shot 2017-04-03 at 11 21 43 am

## Imports: After

screen shot 2017-04-03 at 11 14 47 am

## Sources: Before

screen shot 2017-04-03 at 11 21 51 am

## Sources: After

screen shot 2017-04-03 at 11 20 30 am

### DIFF
--- a/app/views/sources/index.html.haml
+++ b/app/views/sources/index.html.haml
@@ -2,18 +2,16 @@
 
 %h1 Sources
 
-%ul
-  - for source in sources
-    %li= source.name
-
 %table.table
   %thead
     %tr
+      %th &nbsp;
       - for type in ordered_sources.keys
         %th= type.to_s.humanize
   %tbody
-    - (0...Source.count).to_a.each do |position|
+    - (0...Source.count).to_a.each_with_index do |position, index|
       %tr
+        %td #{index + 1}
         - for type in ordered_sources.keys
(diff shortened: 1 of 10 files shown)


app/views/sources/index.html.haml
```diff
@@ -2,18 +2,16 @@
...
-
 %table.table
   %thead
     %tr
+      %th &nbsp;
       - for type in ordered_sources.keys
         %th= type.to_s.humanize
   %tbody
-    - (0...Source.count).to_a.each do |position|
+    - (0...Source.count).to_a.each_with_index do |position, index|
```
anandaroop: Found the iteration logic in this `tbody` a little hard to grok. Couple of observations:

- I think the `.to_a` is unnecessary since a `Range` such as `0...Source.count` is already an `Enumerable` and will respond to `each`

- With a Range like `0...some_num` that starts at zero, `each_with_index{ |value, index| … }` seems superflous, since the value and the index will always be identical. Would a simple `each` suffice here?

gnilekaw:
Thanks for the review @anandaroop

Issue #146, opened 2017-04-03
Add fields for city and country


gnilekaw:
I'm reasonably certain that this will be the first thing people want to search by. 

We could add this to the geocoder job easily.

madeleineb:
Good call. Seems like the Burden city/country data is also pretty clean, and it would be good to be able to import that smoothly.

gnilekaw:
@madeleineb Lmk if you want me to get this in for the mvp

madeleineb:
@gnilekaw We're definitely gonna do it anyway, so if it's fast to add the geocoder job, while not part of that 100% bare-minimum-critical information, it will help the MVP goal of sending useful info on. Let's do it.

PR #150, opened 2017-04-04
Add country and city fields to db and populate with Geocoder

gnilekaw:
Closes https://github.com/artsy/bearden/issues/146


gnilekaw:
Merging

PR #156, opened 2017-04-07
Add sync model

jonallured:
This PR represents my entire week. The memory problems at Heroku were really hard to figure out, but what it seems to have come down to is only loading 1,000 records at a time.

So, this PR introduces a new model called `Sync`. This model does a couple things:

* helps prevent a Sync being started while there's already one running
* stores the state of the syncing operation to that it can be managed

On that last point, the management of a sync, I created a job called `SyncManagementJob` that's run on Heroku every 10 minutes. This job has the responsibility of knowing what needs to be done, which boils down to two things:

* should we start a new sync?
* should we move an exporting sync over to copying?

That first one is easy - if there currently isn't a Sync in progress, then it's time to start one. The second one is a little more complicated, but ends up being a check for state (is the Sync exporting) and a counter check (is `uploaded_parts` equal to `total_parts`).

Ok, so lets talk about a sync operation. Prior to this PR, that work happened in the `DataWarehouseSyncJob`, but that wasn't going to work - I wanted to split it up so that I could create these CSV files in parallel. Three jobs came out of that one:

* StartSyncJob
* OrganizationExportJob
* FinishSyncJob

With the following responsibilities:

## StartSyncJob

* Skip Sync unless there are Imports to sync
* Post to Slack
* Update the states of the Imports to 'syncing'
* Enqueue a bunch of jobs to export all Organizations
* Update the Sync's state and total_pages count

## OrganizationExportJob

* Take in our page number and export to S3 just those Organizations
* Increment `uploaded_parts` counter on Sync

## FinishSyncJob

* Truncate and copy data to Redshift
* Update the states of the Imports to 'synced'
* Post to Slack
* Update the Sync's state to 'finished'

Probably the best part about this PR is that it adds a bunch of tests, haha. The commit that we're improving upon here was never meant to be a long-term solution, so I took the time to wrap some specs around the behavior, but even that could be improved. I don't really have an overall integration test - I'm mostly testing at the unit-level. Open to ideas.

So, did it work? Has this approach solved the memory problems we were having? I'd say so:

screen shot 2017-04-07 at 1 20 44 pm

After the mark at v108 I did two back-to-back syncs with no evidence of memory growth - great success!!

But why?

I don't know, but I think there's a reason the default `in_batches` batch count is 1,000. It seems like when you limit yourself to that many Organizations then the memory is able to be cleaned up properly - load too many and it's not collected the way I would expect it to. Did that convince you?
 Look, I'm not a memory doctor, I'm doing the best I can here.

### Will this work forever? What about when we have 100,000 Organization records?

I'm not sure, but I think so. It *will* mean that we'll end up generating 1,000 CSVs on S3, so that's not ideal, but those COPY statements to Redshift are FAST, so I think it would be ok. As we scale, we'll want to keep an eye on things in case we need to adjust some levers on this process.

/cc @joeyAghion @yuki24


app/jobs/organization_export_job.rb
```diff
@@ -0,0 +1,39 @@
+class OrganizationExportJob < ApplicationJob
+  PART_SIZE = 1_000
```
gnilekaw: We should double check with analytics that it's cool to send Redshift 50 csvs and 50 copy statements. 

@wrgoldstein: Basically, we're dividing the database into chunks of 1000 rows, exporting those as CSVs and then issuing a copy statement to Redshift for each CSV. The lowest number will start at around 50 CSVs and climb as the database grows. Bearden fires this sync off when new datasets are uploaded, so this shouldn't happen _that often_. This is great for Rails performance. And Redshift seems to be fine with it. But I wanted to draw your attention here in case you know of any problems it could cause.

app/models/sync.rb
```diff
@@ -0,0 +1,61 @@
...
+  def finalize
+    machine.trigger SyncMicroMachine::FINALIZE
+  end
+
+  def finish
+    machine.trigger SyncMicroMachine::FINISH
+  end
+
+  # this might not be neccessary
+  def increment_uploaded_parts
```
jonallured: Ugh, totally right - thank you for catching this!! ❤️

app/models/sync.rb
```diff
@@ -0,0 +1,61 @@
...
+    increment! :uploaded_parts
+  end
+
+  def export_folder
+    timestamp = created_at.strftime('%F%T').gsub(/[^0-9a-z ]/i, '')
```
jonallured: Right you are! Your comment actually reminded me of a little ditty we used to do at Hashrocket - here's a post that describes the technique:

http://samuelmullen.com/2009/09/named-datetime-formats-in-ror/

Basically, it's a way of naming these formats so that you don't have to look at something so archaic to get meaning out of it. Also reduces duplication, not that we have a problem with that at the moment.

In any event, I'll get a commit in that adds the initializer.

gnilekaw: %I%%k%n%o%w%%e%x%a%c%t%l%y%%w%h%a%t%%y%o%u%%m%e%a%n

app/jobs/finish_sync_job.rb
```diff
@@ -1,25 +1,29 @@
...
-  def perform
-    return if nothing_to_sync?
+class FinishSyncJob < ApplicationJob
+  attr_accessor :sync
 
-    SlackBot.post('sync starting')
-    update_finished_imports
+  def perform(sync_id)
+    @sync = Sync.find_by id: sync_id
+    return unless sync
```
yuki24: If this is a common pattern, we could add a `rescue_from` to `ApplicationJob`:

```ruby
class ApplicationJob < ActiveJob::Base
  rescue_from ActiveRecord::RecordNotFound do |record_not_found|
    # log if needed
  end
end

class FinishSyncJob < ApplicationJob
  ...

  def perform(sync_id)
    @sync = Sync.find(sync_id) # just find instead of #find_by(...)

    ...
  end
end
```

config/initializers/datetime_formats.rb
```diff
@@ -0,0 +1,6 @@
+{
```
gnilekaw: 👍 Good for humans!

gnilekaw:
🔥

app/jobs/finish_sync_job.rb
jonallured: I like this, but I'm not sure we're ready for it. We aren't logging anything at the moment - we just return from the job. I love that Rails provides this API and when we're ready, I will totally take you up on it! 😉

Issue #152, opened 2017-04-05
Change import name link on Imports page to individual import page


madeleineb:
Clicking on the name of an import now links to the original CSV file. I feel that this should instead take you to the individual import details page—eg. `/imports/5`.

From there, you can also view and download the original file, so, fine to unlink it on the Imports page.

gnilekaw:
Good catch. I totally agree.

Issue #149, opened 2017-04-04
Friendly file upload


madeleineb:
It's hard for humans to remember where files have to be and how to enter a valid URL.

Let's allow any Bearden user to upload a CSV file of any size from their computer easily.

jonallured:
Going to remove the `post-mvp` label from this one per our conversation earlier.

PR #161, opened 2017-04-10
Add upload feature

gnilekaw:
This PR adds a simple file upload to `imports/new` so that users don't have to upload to Dropbox or S3 beforehand. This is good for the user and also better for security, as we can use private files on S3 now.

screen shot 2017-04-10 at 2 22 32 pm

closes https://github.com/artsy/bearden/issues/149
fixes https://github.com/artsy/bearden/issues/152


madeleineb:
🎉 🙌 🎉

db/migrate/20170410135556_add_csv_to_imports.rb
```diff
@@ -0,0 +1,6 @@
+class AddCsvToImports < ActiveRecord::Migration[5.0]
+  def change
+    add_column :imports, :csv, :string
+    remove_column :imports, :uri
```
jonallured: I wonder if renaming would be easier - technically, this migration is not reversible because the `remove_column` doesn't include the type.

One other thought - calling this column `csv` does lock us into a particular format. What was nice about naming the column `uri` is that it didn't imply any particular format. It could just as easily been JSON or XML or whatever.

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,11 @@
+CarrierWave.configure do |config|
+  config.fog_provider = 'fog/aws'
+  config.fog_credentials = {
+    provider: 'AWS',
+    aws_access_key_id: Rails.application.secrets.aws_access_key_id,
+    aws_secret_access_key: Rails.application.secrets.aws_secret_access_key,
+    region: Rails.application.secrets.aws_region
+  }
+  config.fog_directory  = Rails.application.secrets.aws_bucket
+  config.fog_public     = false
```
jonallured: Nicccce!

app/views/imports/index.html.haml
```diff
@@ -15,7 +15,7 @@
       - for import_result in import_results
         %tr
           %td
-            %a{ href: import_result.uri }= import_result.name
+            %a{ href: import_path(import_result.id) }= import_result.name
```
jonallured: I meant to mention this earlier, but forgot - I think this could just as easily be:

```
%td= link_to import_result.name, import_path(import_result.id)
```

app/views/imports/new.html.haml
```diff
@@ -1,6 +1,6 @@
 :coffee
   $ ->
-    $requiredFields = $('#uri, #import_description')
+    $requiredFields = $('#csv, #import_description')
```
jonallured: I'm surprised this works - maybe I don't give Javascript enough credit. 😉

app/views/imports/show.html.haml
```diff
@@ -23,7 +23,7 @@
 %p= import_result.description
 
 %p
-  %a{ href: import.uri }= import.uri
+  %a{ href: import_result.csv.url } Source File
```
jonallured: As above:

```
%p= link_to 'Source File', import_result.csv.url
```

spec/features/csv_import_spec.rb
```diff
@@ -9,7 +9,7 @@
     visit '/imports/new'
 
     select source.name, from: 'Source'
-    fill_in 'File', with: 'spec/fixtures/one_complete_gallery.csv'
+    attach_file('import_csv', 'spec/fixtures/one_complete_gallery.csv')
```
jonallured: Nicccce!

spec/models/raw_input_changes_spec.rb
```diff
@@ -277,10 +277,8 @@
           organization = Fabricate :organization
           Fabricate :website, organization: organization, content: website
 
-          source = Fabricate :source
           import = Fabricate(
             :import,
-            source: source,
```
jonallured: I'm confused - why remove the `source` here?

config/secrets.yml
```diff
@@ -14,10 +14,10 @@ default: &default
   artsy_application_id: <%= ENV['ARTSY_APPLICATION_ID'] %>
   artsy_internal_secret: <%= ENV['ARTSY_INTERNAL_SECRET'] %>
   artsy_application_secret: <%= ENV['ARTSY_APPLICATION_SECRET'] %>
-  aws_access_key_id: <%= ENV['AWS_ACCESS_KEY_ID'] %>
+  aws_access_key_id: <%= ENV['AWS_ACCESS_KEY_ID']  || 'fake-id' %>
   aws_bucket: <%= ENV['AWS_BUCKET'] %>
-  aws_region: <%= ENV['AWS_REGION'] %>
-  aws_secret_access_key: <%= ENV['AWS_SECRET_ACCESS_KEY'] %>
+  aws_region: <%= ENV['AWS_REGION']  || 'fake-region' %>
+  aws_secret_access_key: <%= ENV['AWS_SECRET_ACCESS_KEY']  || 'fake-key' %>
```
jonallured: Rather than adding fallbacks here, I think it might be nice to simply add these values to the `.env.test` file.

spec/features/csv_import_spec.rb
```diff
@@ -2,14 +2,20 @@
 
 feature 'CSV Import' do
   scenario 'Import one complete gallery' do
+    csv_file = 'spec/fixtures/one_complete_gallery.csv'
+
+    allow_any_instance_of(S3CsvUploader).to receive(:store!)
+    allow_any_instance_of(Fog::Storage::AWS::GetObjectUrl)
+      .to receive(:get_object_url).and_return(csv_file)
+
```
jonallured: Oh man, I bet that was so annoying to track down - nice job finding it!

spec/models/raw_input_changes_spec.rb
gnilekaw: I saw that it could be removed because the `:import` fabricator already creates it by default. So, it's redundant, unless you need to test on `source` itself.

config/secrets.yml
gnilekaw: Nice suggestion. Thank you.

spec/models/raw_input_changes_spec.rb
jonallured: facepalm - good call!! ❤️

db/migrate/20170410135556_add_csv_to_imports.rb
gnilekaw: Re `csv`/`uri`: Good point. Both don't seem to capture what's really going on here. See below for what that field looks like. One option is to name it `file`. I'm open to ideas too.

## The raw value 

```sql
select csv from imports limit 1;
                                        csv                                         
------------------------------------------------------------------------------------
 1491845848-17177-0001-6416/burden-complement-_for-import-roster-orgs-condensed.csv
(1 row)
```
## The hydrated value

```ruby
pry(main)> Import.last.csv
  Import Load (0.5ms)  SELECT  "imports".* FROM "imports" ORDER BY "imports"."id" DESC LIMIT $1  [["LIMIT", 1]]
=> #<S3CsvUploader:0x007ff2ead8b410
 @file=
  #<CarrierWave::Storage::Fog::File:0x007ff2ead8ab00
   @base=#<CarrierWave::Storage::Fog:0x007ff2ead8b118 @uploader=#<S3CsvUploader:0x007ff2ead8b410 ...>>,
   @path="uploads/1491848362-20561-0001-2493/_for-import-roster-orgs-condensed.csv",
   @uploader=#<S3CsvUploader:0x007ff2ead8b410 ...>>,
 @model=
  #<Import:0x007ff2eadd2400
   id: 5,
   source_id: 1,
   description: ";';'",
   created_at: Mon, 10 Apr 2017 18:19:22 UTC +00:00,
   updated_at: Mon, 10 Apr 2017 18:19:23 UTC +00:00,
   transformer: "CsvTransformer",
   state: "finalizing",
   csv: "1491848362-20561-0001-2493/_for-import-roster-orgs-condensed.csv">,
 @mounted_as=:csv,
 @storage=
  #<CarrierWave::Storage::Fog:0x007ff2ead8b118 @uploader=#<S3CsvUploader:0x007ff2ead8b410 ...>>,
 @versions={}>
```

jonallured: Wow, trickery! But impressive that they can pack all that info into a string column. 😝 

Yeah, I'm good with `file` or something like that. I think the call site would be nice like that:

```
import.file.url
```

spec/features/csv_import_spec.rb
gnilekaw: It was kind of fun once I it was all over.

db/migrate/20170410135556_add_csv_to_imports.rb
gnilekaw: Yeah, that API works out. If we were to start storing files on the server it would also look good: `File.open(import.file)`. Not that that's on the table!

gnilekaw: Yerg. So using `file` _might_ be starting an infinite loop.

gnilekaw: Everything's cool until you try to call `import.file.url`.

```ruby 
pry(#<ImportsController>)> import
 => #<Import:0x007fe77a49aca0
  id: nil,
  ...

pry(#<ImportsController>)> import.file
 => #<S3CsvUploader:0x007fe77a499800
  @cache_id="1491924610-12378-0001-7992",
  ...

pry(#<ImportsController>)> import.file.url
 SystemStackError: stack level too deep
 from /Users/lancew/.rvm/gems/ruby-2.4.1/gems/carrierwave-1.0.0/lib/carrierwave/uploader/versions.rb:201:in `url'
```

PR #162, opened 2017-04-10
Remove uniqueness constraint from org name

### LINKED ISSUE #104: New lines with matching fields will fail
Take the case of two CSVS:

**A**
```
organization_name,website,location,phone_number,email
"Foo's Bar", example.com, "Detroit, MI", , 
```

**B**
```
organization_name,website,location,phone_number,email
"Foo's Bar", example.com, , , foosbar@example.com
```

Both imports `A` and `B` are equally important. Unfortunately, `B` will fail because the row's import is saved in a transaction and `Foo's Bar` is invalid, since it already exists.

See https://github.com/artsy/bearden/pull/117 for a failing test.

jonallured:
Rather than have this Rankable behave differently than the others, we've
decided that it should rely on Source rank resolution too.

closes #104

### DIFF
--- a/app/models/organization_name.rb
+++ b/app/models/organization_name.rb
@@ -2,7 +2,6 @@ class OrganizationName < ApplicationRecord
   belongs_to :organization
 
   validates :content, presence: true
-  validates_uniqueness_of :content, scope: :organization
 
   has_paper_trail ignore: %i(created_at updated_at)
   include Rankable


app/models/organization_name.rb
```diff
@@ -2,7 +2,6 @@ class OrganizationName < ApplicationRecord
   belongs_to :organization
 
   validates :content, presence: true
-  validates_uniqueness_of :content, scope: :organization
```
gnilekaw: 👋

PR #167, opened 2017-04-11
Handle tag errors on import

jonallured:
Prior to this commit we had implemented passing tag errors up the stack,
but didn't have any specs around it and sure enough, the implementation
wasn't flexible enough. This commit fixes our implementation and also
adds the missing spec.


gnilekaw:
Good catch

PR #168, opened 2017-04-11
Add nav for Syncs 

gnilekaw:
Let's make this public.


jonallured:
💯

madeleineb:
@gnilekaw "syncs" is directing me to `/tags` in staging. Also, cosmetic, but the sign out link is perched awkwardly like a superscript at the right of the last nav item—I'd expect it to align right at the top?

PR #175, opened 2017-04-12
Misc Fixes and Improvements

### LINKED ISSUE #155: Add country and city to CSV and Redshift
Note to self: Don't forget. @jonallured I'm waiting until your work on batches lands before I dig in to this.

### LINKED ISSUE #173: Indicate proper CSV headers in Imports
Would be helpful to humans to have visibility into what headers Bearden expects, and what will happen if their headers don't match. 

bearden import template

gnilekaw:
- Fixes formatting in top nav
- Adds help section for Formatting CSVs
- Adds city and country to converter, transformer, and resolver. 

**NOTE: This PR requires Redshift tables to be updated!**

apr-12-2017 13-14-29

Closes #155
Fixes #173

### DIFF
--- a/app/models/csv_transformer.rb
+++ b/app/models/csv_transformer.rb
@@ -1,7 +1,10 @@
+# rubocop:disable Metrics/MethodLength
 class CsvTransformer
   def self.allowed_headers
     %w[
       email
+      city
+      country
       location
       latitude
       longitude
@@ -41,7 +44,9 @@ class CsvTransformer
 
   def location_attrs
     {
+      city: @data['city'].presence,
       content: @data['location'].presence,
+      country: @data['country'].presence,
       latitude: @data['latitude'].presence,
       longitude: @data['longitude'].presence
     }.compact
--- a/app/views/imports/new.html.haml
+++ b/app/views/imports/new.html.haml
@@ -13,14 +13,12 @@
 
 %h1 Create Import
 
-%ul{ class: %w[nav nav-tabs], role: 'tablist' }
-  %li{ class: 'nav-item'}
-    %a{ class: 'nav-link', href: '#help', data: { toggle: 'tab' }} Help
-  %li{ class: 'nav-item'}
-    %a{ class: %w[nav-link active], href: '#new', data: { toggle: 'tab' }} New
+%ul.nav.nav-tabs{ role: 'tablist' }
+  %li.nav-item= link_to 'Help', '#help', {class: 'nav-link', data: { toggle: 'tab' }}
+  %li.nav-item= link_to 'New', '#new', {class: %w[nav-link active], data: { toggle: 'tab' }}
 
 .tab-content
-  %div{ class: %w[tab-pane], id: 'help' }
+  #help.tab-pane
     %p Bearden expects a CSV with the folowing headers and types of data.
     %p It will skip columns that do not match.
     %table.table
@@ -60,7 +58,7 @@
           %td No
         %tr
 
-  %div{ class: %w[tab-pane active], id: 'new' }
+  #new.tab-pane.active
     %p See the help tab for how to format your CSV
 
     = form_for import do |f|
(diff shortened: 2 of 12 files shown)


app/views/imports/new.html.haml
```diff
@@ -6,16 +6,66 @@
...
+%ul{ class: %w[nav nav-tabs], role: 'tablist' }
+  %li{ class: 'nav-item'}
+    %a{ class: 'nav-link', href: '#help', data: { toggle: 'tab' }} Help
+  %li{ class: 'nav-item'}
+    %a{ class: %w[nav-link active], href: '#new', data: { toggle: 'tab' }} New
+
+.tab-content
+  %div{ class: %w[tab-pane], id: 'help' }
+    %p Bearden expects a CSV with the folowing headers and types of data.
+    %p It will skip columns that do not match.
```
jonallured: Maybe something like this instead:

> It will import, but ignore, columns that do not match.

What I'm wanting to help convey is that we'll take the data even if it doesn't match. Don't know if mine is much clearer, but maybe you see where I'm going with it?

app/views/imports/new.html.haml
```diff
@@ -6,16 +6,66 @@
...
-    = f.collection_select(:source_id, Source.all, :id, :name, {}, { class: 'form-control' })
-  .form-group
-    = f.label :description
-    = f.text_area :description, { class: 'form-control', required: true }
-  = f.submit 'Create', { class: 'btn btn-primary' }
+%ul{ class: %w[nav nav-tabs], role: 'tablist' }
+  %li{ class: 'nav-item'}
+    %a{ class: 'nav-link', href: '#help', data: { toggle: 'tab' }} Help
+  %li{ class: 'nav-item'}
+    %a{ class: %w[nav-link active], href: '#new', data: { toggle: 'tab' }} New
```
jonallured: I think this could be simplified - my rule of thumb is to use the class and id shorthand where possible:

```
%ul.nav.nav-tabs{ role: 'tablist' }
  %li.nav-item= link_to 'Help', '#help', class: 'nav-link', data: { toggle: 'tab' }
  %li.nav-item= link_to 'New', '#new', class: 'nav-link active', data: { toggle: 'tab' }
```

app/views/imports/new.html.haml
```diff
@@ -6,16 +6,66 @@
...
+
+.tab-content
+  %div{ class: %w[tab-pane], id: 'help' }
```
jonallured:
```
#help.tab-pane
```

app/views/imports/new.html.haml
```diff
@@ -6,16 +6,66 @@
...
+          %td <code>email</code>
+          %td contact@artsymail.com
+          %td No
+        %tr
+          %td <code>tag_names</code>
+          %td <a href="/tags" target="_blank">art, photography, gallery</a>
+          %td No
+        %tr
+
+  %div{ class: %w[tab-pane active], id: 'new' }
```
jonallured:
```
#new.tab-pane.active
```

lib/tasks/redshift.rake
```diff
@@ -18,6 +18,8 @@ namespace :redshift do
       conn.exec(
         "CREATE TABLE IF NOT EXISTS #{Redshift::SCHEMA_TABLE} ( \
           bearden_id integer, \
+          city character varying, \
+          country character varying, \
```
jonallured: Awesome - I would have totally forgotten this! Does this change mean that we'll have to re-create the databases on Redshift? When should that occur?

app/models/csv_transformer.rb
```diff
@@ -1,7 +1,10 @@
+# rubocop:disable Metrics/MethodLength
```
jonallured: ugh RuboCop

app/views/imports/new.html.haml
gnilekaw: Good suggestion. This is text from @madeleineb. M., do you have thoughts?

lib/tasks/redshift.rake
gnilekaw: Yes, it does!

PR #177, opened 2017-04-12
Don't include deploy commits in deploy PR

jonallured:
Prior to this commit, the log statement would include the PRs on release
that were for previous deploys. My git-foo wasn't good enough to filter
them, so I just resorted to grep.


gnilekaw:
👍

PR #183, opened 2017-04-13
Convert import data to UTF-8 when possible

### LINKED ISSUE #164: Add pre-flight step for verifying Import file encodings
The `ParseCsvImportJob` needs to have the data it's working with be in UTF-8 format. Add a step to the Import process that does some checks on the Import, one of which is the encoding. While researching this today, I found this library:

https://github.com/brianmario/charlock_holmes

Great library name or greatest?

jonallured:
Rather than forcing the encoding, this commit uses charlock_holmes to
gracefully convert data to UTF-8. I have yet to find data where this
conversion wasn't right, but it's possible we could still have imports
that raise when parsing the CSV.

But, this is much better than we had before and should solve the
majority case.

closes #164


app/jobs/parse_csv_import_job.rb
```diff
@@ -1,17 +1,32 @@
...
+  attr_accessor :import
+
   def perform(import_id)
-    import = Import.find_by id: import_id
-    return unless import
+    @import = Import.find_by id: import_id
+    return unless @import
+
+    create_raw_inputs
+    @import.transform
```
gnilekaw: `transform` is more descriptive up here in `perform`. Feels less lost in the details.

app/jobs/parse_csv_import_job.rb
```diff
@@ -1,17 +1,32 @@
...
 
-    stringio = open(import.file_identifier.url)
-    stringio.set_encoding Encoding::UTF_8
-    data = stringio.read
+  private
+
+  def data
+    raw_data = Faraday.get @import.file_identifier.url
+
+    CharlockHolmes::Converter.convert(
```
gnilekaw: 🕵️

PR #184, opened 2017-04-13
Resolve websites to improve matching logic

### LINKED ISSUE #131: upon import, handle variation in website URLs
- Bearden should recognize all versions of a website—`http://` `https://` `www.` and no `www.`—as the same site. Upon import, websites should match regardless of prefixes.
- Suffixes such as `/` and `/en` should also not create duplicates.

@gnilekaw not part of MVP, since we can handle this offline while prepping CSVs for import for now (find & replace!)

discussed in #82

jonallured:
This commit introduces the concept of resolving a website. We take what
comes from the RawInput and do a little normalization (adding protocol)
and then throw that at Faraday which we've configured to follow
redirects.

Note: we do not perform this resolution if there's a match to begin
with, so we're about as conservative with our network calls as we can
be.

Also note: we blow up the whole row if the website 404s.

closes #131

### DIFF
--- a/app/models/raw_input_changes.rb
+++ b/app/models/raw_input_changes.rb
@@ -1,6 +1,7 @@
 class RawInputChanges
   class InvalidData < StandardError; end
-  class NoWebsiteBuilt < StandardError; end
+
+  VALID_RELATIONS = %i[email location organization_name phone_number].freeze
 
   def self.apply(raw_input)
     new(raw_input).apply
@@ -9,9 +10,7 @@ class RawInputChanges
   def initialize(raw_input)
     @raw_input = raw_input
     @attrs = raw_input.transform
-    @organization = nil
     @relations = {}
-    @relations_to_build = %i[email location organization_name phone_number]
     @error_details = {}
   end
 
@@ -32,41 +31,23 @@ class RawInputChanges
   private
 
   def find_or_create_organization
-    @state = matching_organization ? RawInput::UPDATED : RawInput::CREATED
-    @organization = matching_organization || create_organization
+    url = @attrs.fetch(:website, {})[:content]
+    builder = OrganizationBuilder.new url
+    builder.find_or_create
+    @state = builder.created? ? RawInput::CREATED : RawInput::UPDATED
+    @organization = builder.organization
   end
 
-  def matching_organization
-    matching_website&.organization
-  end
-
-  def matching_website
-    @matching_website ||= Website.find_by(content: url)
-  end
-
-  def url
-    @attrs.fetch(:website, {})[:content]
-  end
-
-  def create_organization
-    @relations_to_build << :website
-    @organization = Organization.create!
+  def relations_to_build
+    VALID_RELATIONS & @attrs.keys
   end
 
   def build_relations
-    @relations_to_build.each do |relation|
-      next unless @attrs[relation]
+    relations_to_build.each do |relation|
       method = relation.to_s.pluralize
       object = @organization.public_send(method).build @attrs[relation]
       @relations[relation] = object
     end
-
-    ensure_website
-  end
-
-  def ensure_website
-    missing_website = @relations_to_build.include?(:website) && @organization.websites.none? # rubocop:disable Metrics/LineLength
-    raise NoWebsiteBuilt if missing_website
   end
 
   def apply_tags
--- a/app/models/website_resolver.rb
+++ b/app/models/website_resolver.rb
@@ -31,7 +31,7 @@ class WebsiteResolver
   end
 
   def protocol?
-    %r{http(s?)://}.match @url
+    @url.start_with? 'http://', 'https://'
   end
 
   def add_protocol
@@ -39,10 +39,10 @@ class WebsiteResolver
   end
 
   def connection
-    options = { callback: method(:callback) }
+    redirect_options = { callback: method(:callback) }
 
     @connection ||= Faraday.new do |faraday|
-      faraday.use FaradayMiddleware::FollowRedirects, options
+      faraday.use FaradayMiddleware::FollowRedirects, redirect_options
       faraday.adapter Faraday.default_adapter
     end
   end
@@ -51,8 +51,8 @@ class WebsiteResolver
     add_result(old)
   end
 
-  def add_result(env)
-    result = { status: env[:status], content: env[:url].to_s }
+  def add_result(faraday_env)
+    result = { status: faraday_env[:status], content: faraday_env[:url].to_s }
     @results << result
   end
 end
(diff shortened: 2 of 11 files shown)


app/models/raw_input_changes.rb
```diff
@@ -41,13 +41,22 @@ def matching_organization
...
+    Website.find_by(content: content)
   end
 
   def url
     @attrs.fetch(:website, {})[:content]
   end
 
+  def resolved_url
+    @resolved_url ||= WebsiteResolver.resolve url
+    @attrs[:website][:content] = @resolved_url if @resolved_url
```
jonallured: I hate side-effects and this is one - how would you know that by calling the `resolved_url` method you'd also be mutating the `@attrs`?? But, I just couldn't think of a way around it - open to ideas!

spec/models/raw_input_changes_spec.rb
```diff
@@ -150,7 +150,7 @@
             organization_name: 'Best Gallery',
             phone_number: '1-800-123-4567',
             tag_names: 'invalid',
-            website: 'examplecom'
+            website: 'example.com'
```
jonallured: With this change, an invalid URL will blow up the whole row before we even attempt any of the other parts of applying the RawInput. If this is a problem, we'd have to refactor even more about how this class works, so I opted for the easy path.

app/models/raw_input_changes.rb
gnilekaw: I've been thinking about this and I feel like we are "losing data," despite keeping the raw input. This mutation could be pointing out a minor flaw in our approach.

If we saved both websites (the `301` and the `200`) we would make fewer network calls when the `301` comes around again. This is the first opportunity we lose out on by not saving both websites to the org.

The second thing we lose out on is harder for me to pin down because it's speculative. In a world where we save both websites, I can imagine the `website` table having a field for `status`, which could have values like `301` and `200`. 

Why save the status? First of all by saving the status we know which website to export (the one that's `200`). Second, saving the status more accurately records our app's progress as it gathers data. Third, in the future, we'll probably be trying to figure out if an org is closed. Going from a `200` to a `404` could be a possible signal. And fourth, for any future crawlers, this status could be super helpful. 

These are just some thoughts I had. Perhaps it complicates things more than needed. What do you think?

spec/models/raw_input_changes_spec.rb
gnilekaw: I could see this causing more work for an upload user if they had a bad website and a bunch of invalid data. But let's hold off and see if it's a problem until we refactor.

jonallured: Cool, glad we agree - fix the problem only if it causes users undue pain. 👍

app/models/raw_input_changes.rb
jonallured: Ok, you've convinced me - I'll work on refactoring this a bit to save both the `301` and the `200`. But I think I'd like to ignore the status part for this PR. If we end up wanting this info, I say we do that in a separate ticket. You good with that?

jonallured:
This PR has gone through some significant change since I first posted it. After some discussion with @gnilekaw, we decided that it made sense to store all the URLs for a given organization and I even added a `status` column to help manage that data. Props to @gnilekaw for strongly encouraging that idea. ❤️ 

In order to write the WebsiteResolver with any type of confidence, I had to write it against actual website redirection, but once it was working the way that I wanted, I had to mock out the network because, as we all know, if your test suite hits the network, then you're doing it wrong:

https://twitter.com/RebeccaSlatkin/status/852627030092939274

Thus, I have introduced Webmock in this PR with some hesitation. More on d945c2a.

Final note: the RawInputChanges class has had some slices cut off of it and is headed in the right direction (I think), but I'm still concerned about how difficult this change was and what that says about the structure of applying RawInputs. Open to other ideas here.

app/models/website_resolver.rb
```diff
@@ -0,0 +1,58 @@
...
+  def resolve_url
+    add_protocol unless protocol?
+    response = connection.get(@url)
+    add_result(response.env)
+  rescue Faraday::Error
+    @results = []
+  end
+
+  def protocol?
+    %r{http(s?)://}.match @url
```
gnilekaw: #minor. Perhaps ditch the regex for `@url.start_with? 'http://', 'https://'`? It's a simple enough regex, but this might be more declarative.

app/models/website_resolver.rb
```diff
@@ -0,0 +1,58 @@
...
+      faraday.use FaradayMiddleware::FollowRedirects, options
+      faraday.adapter Faraday.default_adapter
+    end
+  end
+
+  def callback(old, _)
+    add_result(old)
+  end
+
+  def add_result(env)
```
gnilekaw: I can guess that this argument is a response object, but what does `env` stand for?

app/models/website_resolver.rb
```diff
@@ -0,0 +1,58 @@
...
+  def protocol?
+    %r{http(s?)://}.match @url
+  end
+
+  def add_protocol
+    @url = "http://#{@url}"
+  end
+
+  def connection
+    options = { callback: method(:callback) }
```
gnilekaw: I love this callback/middleware for gathering all redirects. The connection between the two is a little obscure though. What about adding a comment connecting `FollowRedirects` with `options`?

app/models/website_resolver.rb
```diff
@@ -0,0 +1,58 @@
...
+  def resolved_url
+    @results.last&.fetch :content
+  end
+
+  private
+
+  def resolve_url
+    add_protocol unless protocol?
+    response = connection.get(@url)
+    add_result(response.env)
```
gnilekaw: Cool, so this is why [`@results.last`](https://github.com/artsy/bearden/pull/184/commits/2edbb0cfa18739683166bcbb06dcfa0784de53e2#diff-39cf08e9f94dcacae44558329579bafdR20) returns the 200 or 404.

jonallured: Right, we get a callback for each hop except the last one, which is just returned as the response.

app/models/website_resolver.rb
jonallured: OMG, so good - I was trying to find the docs for `start_with?` hoping that maybe it would take more than one argument but couldn't find it. Will change! 👍

app/models/website_resolver.rb
jonallured: It's actually not a response, it's a `Faraday::Env` object. Maybe it would be more clear if I renamed that arg as `faraday_env`?

app/models/website_resolver.rb
jonallured: Or maybe a better name? I'll try calling that local `redirect_options` and see if you like that better.

PR #179, opened 2017-04-13
Add spinner to imports page

gnilekaw:
apr-13-2017 09-39-17

Closes #178

PR #187, opened 2017-04-14
Deploy: convert import data and misc improvements

jonallured:
Included PRs:

* #186
* #185
* #183
* #175
* #179
* #177


jonallured:
Also notable is that this deploy adds an Aptfile which is used by the Heroku apt buildpack so that we can install the binary for charlock_holmes. More in #186.

PR #188, opened 2017-04-14
Fix: Faraday returns a response, not the body

jonallured:
Pretty cool that this was caught because of Sentry and then fixed easily. Pretty not cool that I didn't smoke test this locally before doing my PR. Oh well, mistakes happen! 🤖


gnilekaw:
Thanks!

PR #192, opened 2017-04-14
Export field with array of input sources to Redshift

gnilekaw:
NOTE: This PR changes the Redshift schema.

This PR adds a `sources` field to our CSV export. The goal is to export a new field that contains a list of  all the record's sources, ie, `'Source Foo', 'Source Bar', 'Source Qux'`. These sources can come from an `organization`, a `phone_number`, a `location`, and so on. We provide this list as an easy-ish way for end users to query records which contain a given source.

A new module, `Auditable`, fell out of the refactor. All modules that implement PaperTrail now do so via Auditable. This simplified our models and gave us an added method `model.sources`, which returns all the model's source objects. Previously, these were a pain to call: `organization.emails.first.versions.first.actor.source`. Now you can simply use `organization.emails.first.sources`. Additionally, you can call all an organization's related sources with `organization.contributing_sources`.

One thing that this PR brought up in conversations with @jonallured is, why do we have PaperTrail in the first place when we don't actually edit/update records? 

There is currently no way that an email or a phone number, for instance, would ever be updated. So, what's the value? Well, in a speculative future, if this app adds user CRUD, then it will possibly be possible to update an email or a phone number*. And in that case, our example of an email would have multiple sources, and seeing who changed what will be helpful. This future, however, imagines an "artisanal" dataset, in contrast to our current approach, which is focused more on quantity.

Closes https://github.com/artsy/bearden/issues/182

*Full disclosure: I'm into not ever updating data, instead opting for append-only.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -7,5 +7,26 @@ class Organization < ApplicationRecord
   has_many :tags, through: :organization_tags
   has_many :websites
 
-  has_paper_trail ignore: %i[created_at updated_at]
+  include Auditable
+
+  def contributing_sources
+    auditable_collections.map do |model_name|
+      sources = send(model_name).map(&:sources)
+      sources << self.sources
+    end.flatten.uniq
+  end
+
+  private
+
+  def auditable_collections
+    keys = self.class.reflections.keys
+
+    relations = keys.each_with_object({}) do |key, obj|
+      obj[key] = key.classify.safe_constantize
+    end
+
+    relations.map do |name, klass|
+      name if klass && klass.ancestors.include?(Auditable)
+    end.compact
+  end
 end
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -10,15 +10,16 @@ class Organization < ApplicationRecord
   include Auditable
 
   def contributing_sources
-    auditable_collections.map do |model_name|
-      sources = send(model_name).map(&:sources)
-      sources << self.sources
-    end.flatten.uniq
+    sources = self.sources
+    sources << auditable_relations.map do |model_name|
+      send(model_name).map(&:sources)
+    end
+    sources.flatten.uniq
   end
 
   private
 
-  def auditable_collections
+  def auditable_relations
     keys = self.class.reflections.keys
 
     relations = keys.each_with_object({}) do |key, obj|
--- a/app/models/organization_resolver.rb
+++ b/app/models/organization_resolver.rb
@@ -49,23 +49,8 @@ class OrganizationResolver
     @organization&.tags || []
   end
 
-  def entities
-    [
-      @organization,
-      @email,
-      @location,
-      @organization_name,
-      @phone_number,
-      @website
-    ]
-  end
-
   def source_names
-    entities.reduce([]) do |memo, entity|
-      next entity if entity.nil?
-      name = entity.versions.last.actor.import.source.name
-      memo.include?(name) ? memo : memo << name
-    end
+    @organization.contributing_sources.map(&:source)
   end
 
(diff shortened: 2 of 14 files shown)


app/models/organization_resolver.rb
```diff
@@ -48,6 +49,25 @@ def organization_tag_names
     @organization&.tags || []
   end
 
+  def entities
```
gnilekaw: Not super psyched about this name. I'm open to suggestions!

app/models/organization_resolver.rb
```diff
@@ -48,6 +49,25 @@ def organization_tag_names
...
+      @organization_name,
+      @phone_number,
+      @website
+    ]
+  end
+
+  def source_names
+    entities.reduce([]) do |memo, entity|
+      next entity if entity.nil?
+      name = entity.versions.last.actor.import.source.name
```
gnilekaw: It would be nice to move this up to a higher class. Not exactly sure how that would look. Maybe in ApplicationRecord.

app/models/organization_resolver.rb
```diff
@@ -48,6 +49,25 @@ def organization_tag_names
...
+      memo.include?(name) ? memo : memo << name
+    end
```
starsirius: I might be missing something here, but looks like this function can be simplified as something like:
```ruby
entities.compact.map(&:versions.last.actor.import.source.name).uniq
```

gnilekaw: My thinking is that with `reduce` I only loop once, instead of multiple times. Admittedly, not a big deal with an array with fewer than ten items. If it's a better read for people, I'm happy to do it the way you suggest.

starsirius: It's totally a readability/preference thing so I would say your call.

Regarding the complexity it's not that different with this small `n`, but it's interesting. If I understand that correctly, the complexity of current implementation is
```
1 + 2 + ... + n = n(n - 1) / 2 = O(n^2)
```
because we loop through the `memo` inside of another loop. The implementation above can be `O(n)` I believe.

But, again, readability should be the key here.

jonallured: I have a thought here - I'm thinking the organization should be able to tell us all of the sources that have contributed to it's various rankable data points. One thing that's missing from the implementation in the PR is that only sources of the highest rank are returned, but we want them all.

So yeah, I don't usually make a suggestion like this, but I think we should push this up to the model. 😨 

I think that would be better too because it would open up a sensible space for more unit tests.

Also note, we'd want `versions.first.actor` - def __not__ `last`.

app/models/organization.rb
```diff
@@ -7,5 +7,26 @@ class Organization < ApplicationRecord
...
+  private
+
+  def auditable_collections
+    keys = self.class.reflections.keys
+
+    relations = keys.each_with_object({}) do |key, obj|
+      obj[key] = key.classify.safe_constantize
+    end
+
+    relations.map do |name, klass|
```
gnilekaw: Note: A better option: https://ruby-doc.org/core-2.4.1/Enumerable.html#method-i-flat_map

app/models/organization_resolver.rb
```diff
@@ -48,6 +49,10 @@ def organization_tag_names
     @organization&.tags || []
   end
 
+  def source_names
+    @organization.contributing_sources.map(&:source)
```
gnilekaw: Typo. Should be `:sources`.

jonallured: Hrm, I was going to say that I expected this to be:

```ruby
@organization.contributing_sources.map(&:name)
```

Issue #194, opened 2017-04-18
Overwriting the Redshift schema hoses permissions necessary for Looker


gnilekaw:
```sql
GRANT USAGE
  ON SCHEMA bearden_exports
  TO GROUP read_only;

GRANT SELECT
  ON TABLE bearden_exports.organizations
  TO GROUP read_only;
```

SQL from https://github.com/artsy/collector-experience/issues/202#issuecomment-294508569 and https://github.com/artsy/collector-experience/issues/202#issuecomment-294515228 CC thanks @mikehrom and @wrgoldstein

PR #195, opened 2017-04-18
Fix: grant permissions when creating Redshift db

jonallured:
Took this directly from #194, so hopefully I got the syntax right! ❤️


wrgoldstein:
You shouldn't _need_ to grant usage on the schema more than once unless you're recreating it all the time, but it shouldn't hurt.  Looks good to me.

gnilekaw:
Thanks for adding this! We only run this command when creating a brand new schema or when the altering it, so looks good.

PR #197, opened 2017-04-19
Add in business column

jonallured:
Well, after complaining about the state of `RawInputChanges` in #184, it was actually pretty easy to add this functionality, so maybe things aren't as bad as I feared?

### DIFF
--- a/app/models/csv_converter.rb
+++ b/app/models/csv_converter.rb
@@ -6,6 +6,7 @@ class CsvConverter
       city
       country
       email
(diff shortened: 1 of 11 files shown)


gnilekaw:
Good to go after updating https://github.com/artsy/bearden/blob/master/lib/tasks/redshift.rake#L19. At some point, I'm not saying now, we should rethink how we create schemas. This doing it by hand business will only fly for so long.

PR #198, opened 2017-04-19
Remove Geocoder as Source

jonallured:
Early in the project's life, we assumed Geocoder would be a Source, but
that's not really true.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,8 +5,7 @@ ranks = %i[
 [
   { name: 'Factual', rank: 1 },
   { name: 'HumanOutsourcer', rank: 2 },
-  { name: 'Geocoder', rank: 3 },
-  { name: 'Burden', rank: 4 }
+  { name: 'Burden', rank: 3 }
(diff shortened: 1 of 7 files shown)


db/seeds.rb
```diff
@@ -5,8 +5,7 @@
 [
   { name: 'Factual', rank: 1 },
   { name: 'HumanOutsourcer', rank: 2 },
-  { name: 'Geocoder', rank: 3 },
-  { name: 'Burden', rank: 4 }
+  { name: 'Burden', rank: 3 }
```
gnilekaw: If we do this, we should refactor the Geocoder: https://github.com/artsy/bearden/blob/fb45e09545525aeadabdb95bbe45010128378e04/lib/tasks/geocode.rake#L3

jonallured: Oops, you're totally right. I took a stab at this (refactoring that Job), but I'd like to review it with you later.

gnilekaw: Right on 👌

PR #200, opened 2017-04-19
Fix typo where 'sources' was 'source'

gnilekaw:
As the title suggests, this PR fixes a typo that was preventing pushes to Redshift. It also adds a convenience method, `StartSyncJob.force_sync`, which can be used to force a sync, which is helpful when you change the Redshift schema.


app/jobs/start_sync_job.rb
```diff
@@ -1,12 +1,20 @@
 class StartSyncJob < ApplicationJob
   attr_accessor :sync, :part_size
 
-  def perform(sync_id, part_size = OrganizationExportJob::PART_SIZE)
+  PART_SIZE = OrganizationExportJob::PART_SIZE
+
+  def self.force_sync
+    sync = Sync.create state: SyncMicroMachine::UNSTARTED
```
jonallured: Oops, I forgot that this should actually start the sync in the `SyncMicroMachine::STARTING` state. My bad.

app/jobs/start_sync_job.rb
```diff
@@ -1,12 +1,20 @@
 class StartSyncJob < ApplicationJob
   attr_accessor :sync, :part_size
 
-  def perform(sync_id, part_size = OrganizationExportJob::PART_SIZE)
+  PART_SIZE = OrganizationExportJob::PART_SIZE
```
jonallured: I can't think of a better way to do this, but it's def something I'm going to keep an eye out for.

gnilekaw: One option would be to define it an env var. Batches are configured for performance, right? So unless we're adjusting the batch size at runtime, why define it in the implementation?

jonallured: Whoa, interesting - have one batch size on Heroku and another locally, for instance. That sounds pretty dope!

gnilekaw: Should be a small change. I'll include it here.

app/jobs/organization_export_job.rb
```diff
@@ -1,5 +1,5 @@
 class OrganizationExportJob < ApplicationJob
-  PART_SIZE = 1_000
+  PART_SIZE = ENV['BATCH_EXPORT_SIZE']
```
jonallured: Are you cool with moving this into the `secrets.yml` file? I think we sorta decided that using ENV outside that file was going to be an anti-pattern on this app.

PR #201, opened 2017-04-19
Add permissions to add new and edit existing sources

### LINKED ISSUE #112: Only Bearden Admins can add and edit data sources
While every Artsy Admin user should be able to import and export, only certain people should have the powers to change system-wide things.

## Requirements
- Only certain users are Bearden admins
- Bearden admins can:
     - add data sources #109 
     - Edit rankings of data sources #180 

### Not included, we will still let an engineer do:

- add a new tag #108 
- edit the names of data sources 
- edit a tag's name #111 
- delete a tag (we don't want to make this easy)
- delete a data source (this will make a giant mess)

gnilekaw:
This PR hides view a couple view elements from non-admins so they can't add new sources or edit existing sources.

closes #112


app/controllers/application_controller.rb
```diff
@@ -2,11 +2,13 @@ class ApplicationController < ArtsyAuth::ApplicationController
...
 
-  ALLOWED_ROLES = ['admin'].freeze
+  ADMIN_ROLE = 'marketing_db_admin'.freeze
+  ALLOWED_GRAVITY_ROLES = ['admin', ADMIN_ROLE].freeze
 
   def authorized_artsy_token?(token)
-    roles = user_roles(token)
-    success_or_die(roles)
+    @roles = user_roles(token)
+    @is_admin = admin?
```
starsirius: Looks like we want to make the distinction between gravity admin and bearden admin here. I'd suggest a more explicit naming to avoid confusion.

app/views/layouts/application.html.haml
```diff
@@ -7,17 +7,20 @@
...
-          %ul{ class: %w(navbar-nav mr-auto) }
+        %div{ class: %w[collapse navbar-collapse] }
+          %ul{ class: %w[navbar-nav mr-auto] }
             %li.nav-item= link_to 'Imports', imports_path, { class: nav_class(:imports) }
             %li.nav-item= link_to 'Syncs', syncs_path, { class: nav_class(:syncs) }
             %li.nav-item= link_to 'Sources', sources_path, { class: nav_class(:sources) }
             %li.nav-item= link_to 'Tags', tags_path, { class: nav_class(:tags) }
             %li.nav-item= link_to 'Sign Out', '/sign_out', { class: %w(nav-link) }
 
+      - flash.each do |name, msg|
```
gnilekaw: Not using this any more, but imgunna leave it in

PR #202, opened 2017-04-19
Edit source

### LINKED ISSUE #180: Edit rankings of data sources
Separate edit page for each source

Accessible to Bearden Admins only

@jonallured @gnilekaw

jonallured:
This ended up opening a can of worms on the resolution of source ranks.
What I had was too naive - it only considered the case where you were
adding a new Source. What I had to do was improve this code so that it
supported the case of promoting/demoting existing source ranks.

Luckily, I was able to convince the team to only support editing one
Source at a time, so that means that we have far more simple resolution
logic than we might have otherwise. Even still, this broke my brain for
quite a while!

closes #180

### DIFF
--- a/app/models/source_resolver.rb
+++ b/app/models/source_resolver.rb
@@ -1,34 +1,62 @@
 class SourceResolver
-  def self.resolve(new_source)
-    new(new_source).resolve
+  def self.resolve(adjusted_source)
+    new(adjusted_source).resolve
   end
 
-  def initialize(new_source)
-    @new_source = new_source
+  def initialize(adjusted_source)
+    @adjusted_source = adjusted_source
   end
 
   def resolve
-    return if @new_source.save
+    return if @adjusted_source.save
     adjust_ranks
-    @new_source.save
+    @adjusted_source.save
   end
 
   private
 
+  def conflicted_rank_types
+    @adjusted_source.errors.keys
+  end
+
   def adjust_ranks
     conflicted_rank_types.each do |rank_type|
-      sources_to_adjust(rank_type).each do |source|
-        source.increment!(rank_type)
+      if demotion?(rank_type)
+        move_sources_up(rank_type)
+      else
+        move_sources_down(rank_type)
       end
     end
   end
 
-  def conflicted_rank_types
-    @new_source.errors.keys
+  def demotion?(rank_type)
+    return false unless @adjusted_source.persisted?
+
+    old, new = @adjusted_source.changes[rank_type]
+    old < new
   end
 
-  def sources_to_adjust(rank_type)
-    cutoff = @new_source.rank_for rank_type
-    Source.where("#{rank_type} >= ?", cutoff).order("#{rank_type} DESC")
+  def move_sources_up(rank_type)
+    sources_to_decrement(rank_type).each do |source|
+      source.decrement!(rank_type)
+    end
+  end
+
+  def sources_to_decrement(rank_type)
+    cutoff = @adjusted_source.rank_for rank_type
+    Source.where("#{rank_type} <= ?", cutoff).order(rank_type)
+  end
+
+  def move_sources_down(rank_type)
+    sources_to_increment(rank_type).each do |source|
+      source.increment!(rank_type)
+    end
+  end
+
+  def sources_to_increment(rank_type)
+    cutoff = @adjusted_source.rank_for rank_type
+    old_rank = @adjusted_source.changes[rank_type].first || Source.count + 1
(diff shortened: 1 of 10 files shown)


app/models/source_resolver.rb
```diff
@@ -1,34 +1,62 @@
...
+      source.increment!(rank_type)
+    end
   end
 
-  def sources_to_adjust(rank_type)
-    cutoff = @new_source.rank_for rank_type
-    Source.where("#{rank_type} >= ?", cutoff).order("#{rank_type} DESC")
+  def sources_to_increment(rank_type)
+    cutoff = @adjusted_source.rank_for rank_type
+    old_rank = @adjusted_source.changes[rank_type].first || Source.count + 1
```
gnilekaw: I'm curious about this `||` switch. In what case does `adjusted_source.changes[rank_type]` come up `nil`?

app/views/sources/edit.html.haml
```diff
@@ -0,0 +1,13 @@
...
+
+= form_for source do |f|
+  .form-group
+    = f.label :name
+    = f.text_field :name, { class: 'form-control', required: true }
+
+  - for type in [:email_rank, :location_rank, :organization_name_rank, :phone_number_rank, :website_rank]
+    .form-group
+      = f.label type
+      = f.select(type, source_rank_options(type), {}, { class: 'form-control' })
```
gnilekaw: Q: I'm assuming `source_rank_options` is a helper but I didn't see it in ActionView::Helpers::FormOptionsHelper. Do you know where does it come from?

spec/features/edit_sources_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      organization_name_rank: 2,
+      phone_number_rank: 2,
+      website_rank: 2
+    )
+
+    visit "/sources/#{source.id}/edit"
+
+    fill_in 'Name', with: 'New Name'
+
+    first_option = "1 - insert above #{better_source.name}"
```
gnilekaw: I was confused by this. So ... better source becomes less better source? If I'm reading this correctly, I wonder if we should switch the declarations of `better_source` and `source`. So `better_source` would start out at level `2`, which the reader of this spec would immediately recognize as incorrect vis-a-vis the next source. Then we can imagine our imaginary web user going to this page and saying, "wait, Foo is actually the better source, I should correct this." And at the end of the fable, `better_source` wins and is :1st_place_medal:.

spec/models/source_resolver_spec.rb
```diff
@@ -71,5 +73,37 @@
         expect(second_source.website_rank).to eq 3
       end
     end
+
+    context 'demoting top source' do
+      it 'pulls up other sources' do
+        first_source = Fabricate :source, email_rank: 1
```
gnilekaw: I know it seems backwards, but perhaps the order suggested above could be used here also.

`first_source = Fabricate :source, email_rank: 2`
`second_source = Fabricate :source, email_rank: 1`

Then:

`expect(first_source.reload.email_rank).to eq 1`
`expect(second_source.reload.email_rank).to eq 2`

app/views/sources/edit.html.haml
```diff
@@ -0,0 +1,13 @@
+%h1 Edit Source
+
+= form_for source do |f|
```
jonallured: This should probably be a partial.

app/models/source_resolver.rb
jonallured: In the case where `adjusted_source` is a new source - check it:

```
> existing_source = Source.first
> existing_source.email_rank = 2
=> 2
> existing_source.changes
=> {"email_rank"=>[0, 2]}
> new_source = Source.create email_rank: 3
> new_source.changes
=> {"email_rank"=>[nil, 3]}
```

So, I had to have a fallback for the case where you're adding a new source. In that case you want to move down all sources that follow the rank of the new one. I could have added a third type of `move_sources_BLAH` method, but since this was the only difference, I kept it with the other.

Maybe it's worth pulling out a couple local variables here to name these concepts a little better?

app/views/sources/edit.html.haml
jonallured: Right you are, but you couldn't find it because I made it up! 😃 

https://github.com/artsy/bearden/blob/master/app/helpers/application_helper.rb#L15

See also: c1c4719

spec/features/edit_sources_spec.rb
jonallured: haha, cool - will change!

PR #204, opened 2017-04-24
Refactor to grant admin role based on user id

gnilekaw:
To grant admin to access to edit sources, we're going to hard-code gravity user IDs in `.env`. If this becomes cumbersome, and requires more than minimal changes, it should be refactored so that admin's can grant or revoke access to other admins (or something that we decide when and if the problem presents itself).

See https://github.com/artsy/gravity/pull/10987


gnilekaw:
Pretty cool, right? I just made a guess that it could work. You can also do `BAR={foo: 'bar'}`. But you can't define methods or variables as far as I can see.

PR #209, opened 2017-04-25
Fix import error link

jonallured:
@madeleineb ran into this on production, so this is just a fix.


gnilekaw:
Awesome. Thanks for the quick fix!

madeleineb:
Thanks for the quick fix @jonallured !

PR #218, opened 2017-04-28
Length validations

### LINKED ISSUE #215: Long field values break Redshift
Redshift's `character varying` only allows 256 characters. But Postgres's default for that type is 1GB. After batting this back and forth, Jon and I decided to do two things:

1. Apply sane length validations on fields. 
* Don't send Redshift values with lengths that will break the sync.

jonallured:
Addressing this from two angles:

* ensure that varchar-type fields have sensible limits on their length
* ensure that lists of things (tag and source names) don't exceed limits at export time

closes #215

### DIFF
--- a/app/models/organization_resolver.rb
+++ b/app/models/organization_resolver.rb
@@ -1,4 +1,6 @@
 class OrganizationResolver
+  STRING_LIMIT = 256
+
   def self.resolve(organization)
     new(organization).resolve
   end
@@ -26,7 +28,7 @@ class OrganizationResolver
     rankables.order(created_at: :desc).sort_by(&:rank).first
   end
 
-  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
   def highest_ranked_data
     {
       bearden_id: @organization.id,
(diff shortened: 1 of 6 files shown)


app/models/organization_resolver.rb
```diff
@@ -26,7 +28,7 @@ def find_highest_rankable(rankables)
     rankables.order(created_at: :desc).sort_by(&:rank).first
   end
 
-  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
```
gnilekaw: :shaking_fist: rubbboooocop!

gnilekaw:
These limits are based on Redshift values, but I'm gonna hold my peace ;)

jonallured:
FWIW, I feel your pain @gnilekaw :(

My thought it we start with the validations and the truncations being equal. If we later decide that we want to accept longer inputs, we could change that on the model side and still have the export side limit what's being sent in the CSVs. But I wanted to do that in response to something and this seemed like a good way to get started.

gnilekaw:
Sigh. I agree. The limits might be totally fine, in which case it's the right decision. Let's only revisit if it causes issues with real-world data.

PR #224, opened 2017-05-02
Add '.csv' to files uploaded to S3

gnilekaw:
Seein' if this might fix our dupes bug. I'm skeptical, but at least it'll be an accurate filename.


jonallured:
Thanks for doing this! Maybe it'll solve our problem.

PR #227, opened 2017-05-03
Stop trying to sync when there are PG::Errors

### LINKED ISSUE #216: Syncs should stop when faced with a fatal error from Redshift
screen shot 2017-04-28 at 11 39 57 am

gnilekaw:
When a sync gets a PG error, there's no way that running the sync again is going to succeed the second or _nth_ time. This PR revises the REVERT state to a FAILED state (lolz). The expected behavior is that the PG errors will be returned by `DataWarehouse::Result`, but instead of `REVERT` moving the sync back to `FINISHED`, it moves it to `ERRORS`. The message to slack is then dispatched and the failed sync will not be picked up again until a dev investigates and changes the state.

Fixes #216

### DIFF
--- a/app/jobs/finish_sync_job.rb
+++ b/app/jobs/finish_sync_job.rb
@@ -29,7 +29,7 @@ class FinishSyncJob < ApplicationJob
   end
 
   def update_syncing_imports(success)
-    outcome = success ? :insync : :failed
+    outcome = success ? :insync : :fail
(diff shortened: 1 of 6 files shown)


jonallured:
Cool, good call! I'd like to suggest the following change:

```ruby
self.when(FAIL, SYNCING => FAILED)
```

I think this more closely matches the existing naming pattern.

PR #229, opened 2017-05-05
Use four dynos and four queues to test performance

gnilekaw:
We're hoping to see which job is more of a memory hog


gnilekaw:
Did this wrongly. Closing.

PR #234, opened 2017-05-08
Upgrade rails

jonallured:
I did a `bundle update` so technically this PR also updates some unrelated gems, but the tests are green, so I wasn't going to micro-manage it.

While working on this, I was getting Webmock errors from CarrierWave not behaving and revisited how we were setup on that one. I took a look at the docs and was able to configure things so that we hit the filesystem and that simplified the tests. 👍


gnilekaw:
5.1 🎉

PR #237, opened 2017-05-09
Put exports back on their own queue

jonallured:
These jobs are so slow that they gum up the works for all other job types so we're going to put them back on their own queue.


jonallured:
Doing a quick selfie merge on this to get it on production.

PR #239, opened 2017-05-09
Separate concurrency variables

jonallured:
Prior to this commit it was impossible to change the concurrency on one queue without changing it for all queues.


jonallured:
Another selfie merge to get this to production since it's just a config change.

PR #240, opened 2017-05-09
Use string column for in_business

jonallured:
This commit also includes a temporary rake task to migrate existing data on production.

### DIFF
--- a/lib/tasks/redshift.rake
+++ b/lib/tasks/redshift.rake
@@ -26,7 +26,7 @@ namespace :redshift do
           city character varying, \
           country character varying, \
           email character varying, \
-          in_business boolean, \
+          in_business character varying, \
           latitude double precision, \
           longitude double precision, \
           location character varying, \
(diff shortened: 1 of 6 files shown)


lib/tasks/temporary/in_business_data_migration.rake
```diff
@@ -0,0 +1,15 @@
+desc 'Covert in_business to a string'
```
gnilekaw: This task is in the temporary directory. Are you planning on deleting it? Or is `temporary` a signal that these tasks were run in the past?

lib/tasks/redshift.rake
```diff
@@ -26,7 +26,7 @@ namespace :redshift do
           city character varying, \
           country character varying, \
           email character varying, \
-          in_business boolean, \
+          in_business character varying, \
```
gnilekaw: 👍 

We end up changing the schema often. ... makes me wonder when we should consider automating the drop/create task when it changes.

gnilekaw:
Looks good!

lib/tasks/temporary/in_business_data_migration.rake
jonallured: Yeah, once I've run these data migrations, I'm going to have a PR that removes them. They'll always be in the git history, but I don't want them lingering around.

lib/tasks/redshift.rake
jonallured: I have had a similar thought, but didn't do anything about it. :\

PR #244, opened 2017-05-11
Fix source ranking

gnilekaw:
In the UI, ranking a source last corrupts the hierarchy. Given sources such as `[1, 2, 3]`, moving `1` to `3` results in `[2, 3, 3]`. After reviewing the `source_resolver` and not succeeding in reproducing the bug in tests, I determined the bug was probably in the UI. A failing test confirmed this. 

@jonallured and I decided to change a couple things to in the UI to make things more clear. This PR addresses those revisions, among others.

- [x] Revise language on select options. Currently they say "1 - insert above Foo Source." I'll change this to say "1 - insert below Foo Source." 
- [x] Revise language on selected option to say "2 - Selected rank."
- [x] Add horizontal rule between the name and rank fields to show those are different actions.
- [x] Add the name of the source being edited to the H1.
- [x] Only show "Add to end" option when a new source is being created.
- [ ] On production, restore the source ranks to a correct state.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -15,7 +15,7 @@ module ApplicationHelper
   def source_rank_options(type)
     options = Source.all.map do |source|
       rank = source.rank_for(type)
-      ["#{rank} - insert above #{source.name}", rank]
+      ["#{rank} - insert below #{source.name}", rank]
     end.sort
     final_rank = options.count + 1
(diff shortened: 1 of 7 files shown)


jonallured:
This is great, my only suggestion is to change the action argument on the edit page to 'edit' instead of 'update' and then also changing it in the helper. I think that makes all the language line up a tad nicer.

jonallured:
Thank you for cleaning up my mess! 🙇

PR #245, opened 2017-05-11
Migration tweaks

jonallured:
After some conversations with @gnilekaw, I realized the `in_business` task was just plain wrong. It was also slow, so I improved that as well. While working on this, I realized that our task to clear Sidekiq was inadequate, so I improved that too.

### DIFF
--- a/lib/tasks/clear_sidekiq.rake
+++ b/lib/tasks/clear_sidekiq.rake
@@ -1,14 +1,15 @@
 desc 'Clear all Sidekiq queues'
 task clear_sidekiq: :environment do
-  sets = [
-    Sidekiq::Queue,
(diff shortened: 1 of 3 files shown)


lib/tasks/clear_sidekiq.rake
```diff
@@ -1,14 +1,15 @@
...
+  named_queues = [
+    Sidekiq::ScheduledSet.new,
+    Sidekiq::RetrySet.new,
+    Sidekiq::DeadSet.new
   ]
 
-  sets.each do |set|
-    puts set
-    puts set.new.clear
+  [queues + named_queues].each do |queue|
```
gnilekaw: 👍

PR #246, opened 2017-05-12
Fix sort method for source select options

gnilekaw:
With select options greater than 10 lists appeared thusly `1, 10,  2, 3, ...`. We were calling `sort` on a string value `['10 - Foo Source', 10]`. This PR refactors to sort on the last element of the array, which is an integer.


jonallured:
so good, thank you!! 💯

PR #248, opened 2017-05-16
Add some documentation

### LINKED ISSUE #243: Add documentation
It's about time to add basic documentation! Some kind of data pipeline diagram covering sync and async workflow would be very helpful. Feel free to edit this issue and add anything you'd like.

jonallured:
I took a stab at some documentation for Bearden. It's probably easiest to read this on my fork here:

https://github.com/jonallured/bearden/blob/docs/docs/README.md

While working on this, I decided to see if I could use graphviz to create some graphs, especially around how the states of an Import and Sync work. Then I had the idea to create a task that could graph an organization, so that's in here too although technically it's not really documentation.

Open to ideas about what else should be covered here!

closes #243


gnilekaw:
[I'm going to add revisions here as I come across them]

Under Import Workers paragraph, change "fall" to "call":
> The last of these recursive jobs will _call_ finalize

docs/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Import Workers
+
+Importing a CSV file into Bearden is accomplished by workers. This process
+starts in the [ImportsController][imports_controller], when it [calls
+`parse`][import_parse] on the `Import` record. That, in turn, enqueues a
+`ParseCsvImportJob`. That job creates `RawInput` records for each row in the CSV
+file and then when it's done doing that kicks off a recursive job to apply each
+in turn using the `RawInputTransformJob`. The last of these recursive jobs will
+fall `finalize` on the `Import` which enqueues the `FinalizeImportJob`. That job
```
gnilekaw: Should be "call" not "fall"

docs/README.md
```diff
@@ -0,0 +1,142 @@
...
+[sync_mm]: /app/models/sync_micro_machine.rb
+[sync_graph]: /docs/SyncMicroMachine_graph.png
+
+### Sync Workers
+
+Exporting data to Redshift is accomplished by workers. Every 10 minutes when the
+sync task is run it enqueues a [`SyncManagementJob`][sync_mgmt] which does one
+of these things:
+
+* create a new `Sync`
```
gnilekaw: Verbs in list should be plural: "creates" and "kicks off."

docs/README.md
```diff
@@ -0,0 +1,142 @@
...
+[org_resolver]: /app/models/organization_resolver.rb
+[sync_copy]: /app/models/sync.rb#L25
+[sync_finish_job]: /app/jobs/finish_sync_job.rb
+[data_warehouse]: /app/models/data_warehouse.rb
+
+## Resolving Organizations
+
+Yes Bearden imports and exports, but the thing that makes it special is the way
+that it models data and resolves `Organization` records. Rather than simply
+having an email column on the organizations table, for example, we have table of
```
gnilekaw: Add "a" to "we have _a_ table of..."

docs/README.md
```diff
@@ -0,0 +1,142 @@
...
+email addresses that point to their organization. This means the actual
+`Organizaiton` has very little on it and serves mostly as a way to relate all
+the details of an entity. Each of these details points back to the `Source` that
+created it, which means that we can always know which details are most accurate.
```
gnilekaw: Possibly consider revising or cutting,
> which means that we can always know which details are most accurate 

On its own, pointing back to the source doesn't create accuracy, but it allows us to know where the data came from, which is a subtle difference.

What would you think about adding in a brief section that expands on `Source` and ranking? The source/rank is what allows the user to model a hierarchy of dependable sources. That's sort of the beginning and ending of the story of Bearden, right?

docs/README.md
```diff
@@ -0,0 +1,142 @@
...
+[sync_finish_job]: /app/jobs/finish_sync_job.rb
+[data_warehouse]: /app/models/data_warehouse.rb
+
+## Resolving Organizations
+
+Yes Bearden imports and exports, but the thing that makes it special is the way
+that it models data and resolves `Organization` records. Rather than simply
+having an email column on the organizations table, for example, we have table of
+email addresses that point to their organization. This means the actual
+`Organizaiton` has very little on it and serves mostly as a way to relate all
```
gnilekaw: Spelling for "Organization."

docs/README.md
```diff
@@ -0,0 +1,142 @@
...
+Yes Bearden imports and exports, but the thing that makes it special is the way
+that it models data and resolves `Organization` records. Rather than simply
+having an email column on the organizations table, for example, we have a table
+of email addresses that point to their organization. This means the actual
+`Organization` has very little on it and serves mostly as a way to relate all
+the details of an entity. Each of these details points back to the `Source` that
+created it, which means that we always know which details are most accurate.
+
+Evaluating this accuracy happens at export time, during the process of syncing
+with Redshift. Waiting until this point in the process means means that we
```
starsirius: "...means means..."

docs/README.md
```diff
@@ -0,0 +1,142 @@
...
+then from there walk back up to the `Source` that has a rank.
+
+[rankable]: app/models/rankable.rb
+
+### The Rules
+
+The resolution rules are:
+
+* use the `Rankable` with the lowest rank (1 is better than 2)
+* break rank ties with `created_at` - newer is better
```
starsirius: #minor. Maybe keep the format consistent:

... with `created_at` (newer is better)

docs/README.md
```diff
@@ -0,0 +1,142 @@
...
+follow along with the progress of a `Sync`, you could occasionally refresh the
+[sync list page][sync_list]. The process is done in batches driven by an ENV var
+called `BATCH_EXPORT_SIZE`, currently set to 1,000. Each batch of records is
+resolved and then written to its own file on S3. When all batches have finished,
+we truncate the table on Redshift and copy over the data.
+
+Note: Sync seemed like a good word here, but it implies a two-way flow of data
+and that's not what we have. Bearden writes data to Redshift, but doesn't read
+anything. Another word might have been better, but sync is how we all talked
+about it. Language is weird.
```
madeleineb: Great note 😸

docs/README.md
```diff
@@ -0,0 +1,142 @@
...
+having an email column on the organizations table, for example, we have a table
+of email addresses that point to their organization. This means the actual
+`Organization` has very little on it and serves mostly as a way to relate all
+the details of an entity. Each of these details points back to the `Source` that
+created it, which means that we always know which details are most accurate.
+
+Evaluating this accuracy happens at export time, during the process of syncing
+with Redshift. Waiting until this point in the process means means that we
+delay the flattening to as late as possible and reflect the most up-to-date
+information as possible.
```
madeleineb: rather than "up-to-date", "information from the highest-ranked source." This, or below in Rankables, might be a good place to introduce the concept of each Source having a rank for each field, which is alluded to elsewhere but not explicitly stated. The per-field aspect also makes Bearden special.

starsirius:
This is an awesome start! The state machine diagrams are super helpful. Thanks, @jonallured! 💯 

A few thoughts I have:

* Since we have the ERD generated, it may worth just embedding it in the readme to capture the static side of Bearden.
* The paragraphs describing the data flow is clear. Another option is to capture the dynamic part of the flow with some diagrams, like Sequence Diagram, and visualize it. I was trying to do similar things to [visualize cross-app interaction](https://github.com/starsirius/volt/pull/1) in Volt, given the complexity involved. I can see that being useful to capture data pipeline as well as internal components/tasks. However, we should only create those diagrams when needed.

docs/graphs/org-email.dot
```diff
@@ -0,0 +1,9 @@
+digraph g {
```
gnilekaw: To draw this you can use `dot -Tpng docs/graphs/org-email.dot -O`

docs/README.md
```diff
@@ -109,20 +109,32 @@ Redshift.
...
+
+Each of these details, such as an email, location, phone number and so on,
+points back to the `Source` that created it. Consider the following graph.
+
+![Organization and email graph][org_email_graph]
+
+Given multiple fields, an organization resolves to the highest ranked source for
+that field. Since `email_2` originates from the higher-ranked `source_b`, the
+output to Redshift will only contain that `email_2`. The lesser-ranked values
+are ignored.
```
jonallured: Awesome graph - great idea! My only thought is keeping language consistent regarding higher and lower ranks. In the section about the rules, I wrote that lower rank is better, so maybe flip the language here?

docs/README.md
jonallured: Great feedback, adding a section for the way we split rank into multiple fields. 👍

jonallured:
Awesome feedback everyone - thanks for taking the time to do that!!

I've pushed a couple more commits that address most of it, but if I missed something please let me know!

@starsirius I did add a mention of the ERD - embedding it is a little tricky because it's such a wide image. I think it's better for someone to open it on their machine so they can zoom in or whatever. As far as sequence diagrams go, I'm not a big fan because I find them hard to follow, but if you want to add one, that would be great!

@madeleineb I did add the section you suggested - let me know if that's what you were thinking.

I just have one outstanding item for @gnilekaw and it sounds like he's going to address it later today, so I'm thinking once that lands I'm going to merge this and call it good.

Not that documentation is ever actually done, but I think this PR is a good start and ready to ship! ❤️

madeleineb:
@jonallured Awesome, thanks for adding that section about sources! Puts it succinctly and clearly 👍 I think it's fine to use "lowest" to describe source with rank no. 1 if that's what makes sense to most people—1 is the smallest number but calling it lowest is a little counterintuitive for me, because I think about 1 being "most authoritative" and therefore "highest", but that might just be me. Anyhow, so long as it's consistent, whatever will be clearer to devs, since that's who this documentation will serve!

spec/truncation_helper.rb
```diff
@@ -1,13 +0,0 @@
-require 'rails_helper'
```
gnilekaw: truncation helper been truncated

PR #249, opened 2017-05-16
Upgrade Rails and Bundle Update

jonallured:


gnilekaw:
Thanks for keeping things up to date

PR #256, opened 2017-05-22
Source rank migration

gnilekaw:
Reorders the ranks hierarchy. In data we have corrupted rank hierarchies, such as `[1,2,3,4,5,7,8,9,10,11,12,13,13]`. 

This migration will renumber ranks with a new hierarchy.  In the case above, starting after `5`, each item will be renumbered: `7` => `6`, `8` => `9`. 

Notice there is a tie in the last two items, `13` and `13`. How do we know which one comes first? Luckily it doesn't matter. If the rank is lower than one, then the source won't be selected for export, so we can safely ignore this detail.

To see the migration it's more clean [here](https://github.com/artsy/bearden/pull/256/commits/15417ff474dff5514618aea84de9863092d543b9). The other commit adds a namespace for other rake tasks.

### DIFF
--- a/lib/tasks/temporary/in_business_data_migration.rake
+++ b/lib/tasks/temporary/in_business_data_migration.rake
@@ -1,17 +1,19 @@
-desc 'Convert in_business to a string'
-task in_business_data_migration: :environment do
-  PaperTrail.track_changes_with('data migration') do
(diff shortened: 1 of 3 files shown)


jonallured:
Nice one!

PR #257, opened 2017-05-24
Bump location redshift field to 1024 chars

gnilekaw:
Syncs were failing because Arabic characters were getting counted as bytes, which quickly exceeded the former limit of 256.


jonallured:
Makes sense to me - merging!

Issue #258, opened 2017-05-26
Be less picky when matching on website to prevent duplicates


madeleineb:
In two different Bearden records (114389 and 34957), `http://www.borsini-burr.com` has a status of 301 in one record and 200 in another. 

cc: @gnilekaw

PR #260, opened 2017-05-31
Misc maintainance

gnilekaw:
Several small changes to address issues that arose during the last import, namely:

- Don't trigger a sync if all of the imported CSVs contain 100% errors
- Allow mixed-case strings in CSV tag names
- Add a failing (but skipped test) for a known bug which is creating duplicates (see #258)
- Add a new rake task `rake report:duplicate_organizations` to print out a count and a list of duplicate organizations

### DIFF
--- a/app/jobs/start_sync_job.rb
+++ b/app/jobs/start_sync_job.rb
@@ -35,7 +35,7 @@ class StartSyncJob < ApplicationJob
       imports_to_sync -= 1
     end
 
-    return imports_to_sync > 0
+    imports_to_sync.positive?
   end
 
   def start_sync
--- a/app/jobs/start_sync_job.rb
+++ b/app/jobs/start_sync_job.rb
@@ -13,7 +13,7 @@ class StartSyncJob < ApplicationJob
 
     return unless sync
 
-    if force || imports_to_sync?
+    if force || should_sync?
       start_sync
     else
       sync.skip
@@ -22,8 +22,20 @@ class StartSyncJob < ApplicationJob
 
   private
 
-  def imports_to_sync?
-    Import.where(state: ImportMicroMachine::FINISHED).any?
+  def should_sync?
+    finished_imports = Import.where(state: ImportMicroMachine::FINISHED)
+
+    return false unless finished_imports
+
+    imports_to_sync = finished_imports.count
+
+    finished_imports.each do |import|
+      errors_count = import.raw_inputs.where('exception is not null').count
+      next unless import.raw_inputs.count == errors_count
+      imports_to_sync -= 1
+    end
+
+    return imports_to_sync > 0
   end
 
   def start_sync
--- a/app/jobs/start_sync_job.rb
+++ b/app/jobs/start_sync_job.rb
@@ -13,7 +13,7 @@ class StartSyncJob < ApplicationJob
 
     return unless sync
 
-    if force || should_sync?
+    if force || some_imports_should_sync?
       start_sync
     else
       sync.skip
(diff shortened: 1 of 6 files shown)


app/jobs/start_sync_job.rb
```diff
@@ -22,8 +22,20 @@ def perform(sync_id, force: false)
...
   private
 
-  def imports_to_sync?
-    Import.where(state: ImportMicroMachine::FINISHED).any?
+  def should_sync?
+    finished_imports = Import.where(state: ImportMicroMachine::FINISHED)
+
+    return false unless finished_imports
+
+    imports_to_sync = finished_imports.count
```
gnilekaw: We start with `n` number of imports. Most likely it's one, but could be more.

app/jobs/start_sync_job.rb
```diff
@@ -22,8 +22,20 @@ def perform(sync_id, force: false)
...
+
+    finished_imports.each do |import|
+      errors_count = import.raw_inputs.where('exception is not null').count
+      next unless import.raw_inputs.count == errors_count
+      imports_to_sync -= 1
```
gnilekaw: Then, if an import has an equal number of errors as it does rows, then we subtract one from the number of imports to sync.

app/jobs/start_sync_job.rb
```diff
@@ -22,8 +22,20 @@ def perform(sync_id, force: false)
...
+    end
+
+    imports_to_sync.positive?
```
gnilekaw: If that's less than one, then there are no imports to sync, and it returns false.

lib/tasks/report_duplicate_organizations.rake
```diff
@@ -0,0 +1,17 @@
...
+    end
+
+    duplicates = domains.select do |_website, grouped_results|
+      next false unless grouped_results.size > 1
+      grouped_results.map(&:last).uniq.size > 1
+    end
+
+    puts "Count: #{duplicates.values.map(&:last).flatten.uniq.count}"
+    puts duplicates
+  end
```
gnilekaw: Changed the names around here, but it's otherwise the same from yesterday when we wrote it.

spec/jobs/start_sync_job_spec.rb
```diff
@@ -15,13 +15,23 @@
     context 'with an import to sync' do
       it 'posts to Slack, updates the sync and enqueues export jobs' do
         sync = Fabricate :sync, state: SyncMicroMachine::STARTING
-        Fabricate :import, state: ImportMicroMachine::FINISHED
-        Fabricate.times 3, :organization
```
gnilekaw: Fabricating three orgs here is superfluous.

app/jobs/start_sync_job.rb
jonallured: Cool, I like this! My only thought is to move the knowledge around a little. What if we added a method to Import called `every_raw_inputs_failed`:

```ruby
class Import
  def every_raw_inputs_failed?
    raw_inputs.where.not(exception: nil).count == raw_inputs.count
  end
end
```

Then, the worker could just do something like this:

```ruby
def should_sync?
  finished_imports = Import.where(state: ImportMicroMachine::FINISHED)
  finished_imports.all?(&:every_raw_inputs_failed?)
end
```

I realize `every_raw_inputs_failed?` is pretty long, def open to ideas there.

I think a refactor like this is nice so that we can keep the worker ignorant about the details of what makes an import something that doesn't need to be synced.

jonallured: Oh, maybe that's better - what if this method on Import was called `should_sync?`?

lib/tasks/report_duplicate_organizations.rake
jonallured: 💯

spec/jobs/start_sync_job_spec.rb
jonallured: Oops!

app/jobs/start_sync_job.rb
gnilekaw: These are excellent suggestions. See revisions.

jonallured: Noice!

PR #261, opened 2017-06-01
Fail CI builds when migration errors

jonallured:
This is something I picked up from @ashkan18 on Impulse. When you pass the `exit-code` flag and there are errors in migrations then the build will fail rather than silently failing. Seems like a best practice for any of our apps that use Heroku and Postgres.


gnilekaw:
👍

PR #266, opened 2017-06-20
Upgrade artsy auth

jonallured:
This PR upgrades us to the latest artsy-auth gem version. That upgrade switches from inheritance to composition, but the contract is really the same - implement `authorized_artsy_token?` and if you return false then the user will get a 401.

What I noticed while doing this upgrade was that rather than taking this behavior, we were jumping in with a 404. I don't know if this was intentional, but I think it's just as easy to take the default behavior and return the 401. For APIs or customer-facing apps, I would care a lot more about leaking information, but for this app it seems 💯 to me.


app/controllers/application_controller.rb
```diff
@@ -24,6 +23,7 @@ def user(token)
   def admin?
     ADMIN_USERS.member? @user[:uid]
   end
+  helper_method :admin?
```
gnilekaw: 👍

app/controllers/application_controller.rb
```diff
@@ -31,6 +23,16 @@ def ssl_configured?
     !Rails.env.development?
   end
 
+  def decode_user(token)
```
gnilekaw: Yes, this is more clear.

PR #270, opened 2017-06-21
Switch to Circle 2.0

jonallured:
Experimenting with CircleCI 2.0.


jonallured:
Well it only took me about 30 builds to get this to work, but I did it. I'm not that much a fan of the new way configuration works, but you can't argue that users now have a lot more control! I guess I just miss the way Circle embraced convention over configuration. :\

jonallured:
Merging this in order to babysit deployments.

PR #272, opened 2017-06-22
Circle 2.0 Upgrade -- Final

jonallured:
UGGGH


orta:
You got this ✋

jonallured:
Update: I've been in contact with Circle support. The problem, as far as I can tell, is that the $HEROKU_LOGIN and $HEROKU_API_KEY vars aren't being set. I confirmed this by SSHing into a build and just echoing them. I also echoed the `~/.netrc` file, which didn't have those values set as it should. Support said that my config looked good, so I've asked them to explain why those vars aren't set. Sigh.

jonallured:
Circle Support and I finally figured this out! I was attempting to use env vars in a forked branch - can't do that with the "Pass secrets to builds from forked pull requests" set to off! Basically, those vars aren't set for PRs, to protect the sensitivity of our Heroku API key, which is a good thing! But everyone involved forgot about this. So, once Support mentioned this, I pushed this work to a branch on upstream, saw that it was working there and then updated the PR. Once I merge this work, then all will be well with the world and I can forget this ever happened!! 😵

PR #271, opened 2017-06-22
Fix deployment config for Circle 2.0

jonallured:
This was done following this walkthrough:

https://circleci.com/docs/2.0/project-walkthrough/#deploying-to-heroku

I ended up generating a new SSH keypair and storing those creds in 1Password for posterity.

PR #273, opened 2017-06-26
Downgrade Circle

jonallured:
I can't get Circle 2.0 to properly deploy to Heroku, reverting this stuff for now.

From the `downgrade-circle` branch:

```
% git diff staging/master
%
```

So that should mean that this PR simply reverts #270 and #271.


jonallured:
Doing a selfie-merge and then going to babysit the deploy to ensure the revert worked as I expect.

Issue #263, opened 2017-06-02
Migrate country, city, lat and long from Factual imports into Bearden records


madeleineb:

gnilekaw:
This can be closed. Bearden supports these fields https://github.com/artsy/bearden/blob/master/app/models/csv_transformer.rb#L5-L16

PR #274, opened 2017-06-26
Migrate cities for Factual imports

gnilekaw:
The cities didn't get migrated with factual, because that field is called `locality` (versus `city`) in the CSV. This migration sorts through those raw input records and adds city fields that are blank to existing locations.

Closes https://github.com/artsy/collector-experience/issues/353
Closes https://github.com/artsy/bearden/issues/263

### DIFF
--- a/lib/tasks/temporary/factual_migration.rake
+++ b/lib/tasks/temporary/factual_migration.rake
@@ -9,12 +9,13 @@ namespace :temporary do
       raw_inputs.each do |raw_input|
         data = raw_input.data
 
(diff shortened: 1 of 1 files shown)


jonallured:
This looks good to me! My only thought was about `whodunnit` - do you think it makes sense to set this to some value that will reflect that a rake task was the reason for the addition of the city? Thinking about that audit trail. 🗒 🖊

gnilekaw:
Good call. Adding a whodunnit named `PR#274`.

PR #275, opened 2017-06-26
Deploy

gnilekaw:
:shipit:


gnilekaw:
OK. I'm going to run this migration on production now. The script is quite slow and sometimes Heroku hangs up. But it's a one-time thing so :man_shrugging:. 

To test whether this if successful I select the factual locations that don't have a city but do have coordinates. This is ~17,800 records. 

```ruby
factual_locations = Location.where(city: nil).where('longitude > 0').where('latitude > 0').select do |location| 
  location&.sources.pluck(:name).any? { |n| n == 'Factual'}
end

factual_locations.count
```

Then after running the script, this count should be 0.

```ruby
factual_locations.select{|l| l&.actors.pluck(:data).any?{|d| d['locality'].present?}}.count
```

gnilekaw:
All right! This is done. This morning I ran the job one final time. Before doing so, I turned off scheduled jobs. I think there were some issues with database locks ... hard to say for sure, unfortunately. Jobs are scheduled again and maintenance mode is off.

PR #276, opened 2017-06-30
Add field for Locations.geocode_response

gnilekaw:
This PR follows up on https://github.com/artsy/bearden/pull/274, because, as @madeleineb pointed out, around ~38k records on Redshift still seem to be not geocoded. 

Looking deeper into the data, there are a lot of records that are marked as geocoded (they have non-nil latitudes and longitudes), but do not have meaningful data (nil city and country fields).

Unfortunately, checking for non-nil latitudes and longitudes does not return an accurate picture of whether or not a location has been geocoded. This is because, for imports, we allow users to include `latitude` and `longitude` fields. That means locations can have blank city or country fields and never be found by the geocoder. 

Other things might also be at play here. More below on that.

### TL;DR

My solution is to add more visibility into geocoder by saving the response in a new field, `Locations.geocode_response`. This will allow us to return to some of these locations for better analysis after geocoder has been running. 

What does that mean for current locations? It does mean that zero of the 150,518 locations will be marked as geocoded. Geocoder will run daily, as it does now, making requests every `0.2.seconds`. 

From the perspective of Looker/Redshift, not much will change. When new data is synced, there will simply be more accurate location records. There will be no loss of data.

### More background:

This query selects all the locations that we humans consider to be not geocoded. There are 70,794 such locations.

```sql
 select id, latitude, longitude, city, country, left(content, 15) from locations where latitude = 0 or longitude = 0 or city is null or country is null;
```

There are no locations such that latitude or longitude are null. This means that, according to the (flawed) geocoder, all locations have been geocoded. 

So why do some results seem to be not geocoded? Here are some examples from looking at the data.

- `Location.find(77)` is a location that has fallback lat and long (`[0,0]`). This suggests the geocode results returned nil. Yet `geoTest(77)` shows that city and country should be `["Innsbruck", "Austria”]`. It's uncertain why this is happening. The bright side is that locations such as this should be fixed by running the geocoder on them again.
- `Location.find(4700)` shows a location with `content: "US, Washington”`. After geocoding, this resolves to `country: United States and city: nil`. This looks like a missing city, but it's accurate. I saw many locations like this one.
- `Location.find(4418)` is a location with `content: “JP, Tokyo”`. It is geocoded with `country: “Japan”, city: nil`. It looks like an error, and yet these are the results given by geocoder. 
- Another thing to consider is that `rankable` selects the top location based on what a human thinks is the best bulk source of data, not the best actual location. This could result in some amount of errors. Probably not the bulk of them though.

```ruby
# Test geocode results for a location
def geoTest(id)
  r = Geocoder.search(Location.find(id).content).first
  [r.city, r.country]
end
```

### DIFF
--- a/app/jobs/geocode_location_job.rb
+++ b/app/jobs/geocode_location_job.rb
@@ -1,4 +1,6 @@
 class GeocodeLocationJob < ApplicationJob
+  class NoResultsFound < StandardError; end
+
   JOB_DELAY = 0.2.seconds
   FALLBACK_COORDINATES = [0.0, 0.0].freeze
   FALLBACK_COUNTRY = nil
@@ -7,7 +9,7 @@ class GeocodeLocationJob < ApplicationJob
   def perform
     location = next_location
     return unless location
-    geocode(location) if location&.geocodable?
+    geocode(location)
     enqueue_next_job
   end
 
@@ -29,7 +31,8 @@ class GeocodeLocationJob < ApplicationJob
       latitude: result.coordinates[0],
       longitude: result.coordinates[1],
       country: result.country,
-      city: result.city
+      city: result.city,
+      geocode_response: result
     }
   end
 
@@ -38,7 +41,8 @@ class GeocodeLocationJob < ApplicationJob
       latitude: FALLBACK_COORDINATES[0],
       longitude: FALLBACK_COORDINATES[1],
       country: FALLBACK_COUNTRY,
-      city: FALLBACK_CITY
+      city: FALLBACK_CITY,
+      geocode_response: NoResultsFound.to_s
     }
   end
 
@@ -47,6 +51,6 @@ class GeocodeLocationJob < ApplicationJob
   end
 
   def next_location
-    Location.where('latitude IS NULL OR longitude IS NULL').first
+    Location.where(geocode_response: nil).first
   end
 end
--- a/app/jobs/geocode_location_job.rb
+++ b/app/jobs/geocode_location_job.rb
@@ -1,10 +1,9 @@
 class GeocodeLocationJob < ApplicationJob
-  class NoResultsFound < StandardError; end
-
-  JOB_DELAY = 0.2.seconds
   FALLBACK_COORDINATES = [0.0, 0.0].freeze
   FALLBACK_COUNTRY = nil
   FALLBACK_CITY = nil
+  JOB_DELAY = 0.2.seconds
+  NO_RESULTS_FOUND = 'NO_RESULTS_FOUND'.freeze
 
   def perform
     location = next_location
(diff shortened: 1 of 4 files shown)


app/jobs/geocode_location_job.rb
```diff
@@ -1,4 +1,6 @@
 class GeocodeLocationJob < ApplicationJob
+  class NoResultsFound < StandardError; end
```
anandaroop: It seems funny to me to have an `Error` class that never gets thrown, but I do see how this gets the job done below…

app/jobs/geocode_location_job.rb
```diff
@@ -47,6 +51,6 @@ def enqueue_next_job
   end
 
   def next_location
-    Location.where('latitude IS NULL OR longitude IS NULL').first
+    Location.where(geocode_response: nil).first
```
anandaroop: My concern here is that not only yet-to-be-geocoded `Locations`, but also failed-to-be-geocoded `Locations` would result in a `nil` response, and thus be selected over and over for geocoding. 

In the most pathological case you'd have one un-geocoded `Location`, with a typo or otherwise un-geocodable `content`, being attempted for geocoding every 200 ms.

Does that ring true to you? If so I wonder if maybe a simple `geocoded_at` timestamp would be preferable to `geocode_response`. (That would also eliminate the need for the custom Error above.)

But, for better and for worse, that would eliminate the sort of "automatic retry" behavior you'd be getting with this PR's implementation.

gnilekaw: Hmm. How do you see the automatic retry behavior starting? I was planning on failed geocode responses not being nil, because they would have this error: https://github.com/artsy/bearden/pull/276/files#diff-462e4ddf943cd6a3904178ecc6528df4R45.

anandaroop: Gotcha, so `geocode_response` is never actually `nil` because it is explicitly set to that error string whenever there is no match at all from `Geocoder.search`. Ok that makes sense — thanks for clarifying the fallback logic.

gnilekaw: I thought I'd missed something! I see your point about creating a custom error and then not throwing or catching it. I could use a string or another value instead. What do you think?

jonallured: Yeah, I think a constant might be a good solution here. Rather than having an exception class, maybe a string value to reflect the conditions? Would it be worth putting a `not null` on that column?

gnilekaw: 👍 for using a constant. 

I don't think we could use `not null` because to select a non-geocoded location it uses `Location.where(geocode_response: nil).first`.

gnilekaw: @anandaroop `geocode_response` is `nil` when the location has not been geocoded. But, once selected by the geocoder, there are only two code paths: error or result.

anandaroop: That makes sense. `Locations` first get instantiated with a null `geocode_response`, which sounds right.

jonallured:
Looking at this again, I think this is ready to merge - last call for comments! /cc @anandaroop

anandaroop:
👍 LGTM

PR #284, opened 2017-07-13
Onboarding tweaks

anandaroop:
So Lance and I paired a couple of weeks ago on Bearden and in the process I found a few minor wrinkles in the dev on-boarding process.

Thought I'd finally PR these and see if they look good to you.


db/seeds.rb
```diff
@@ -1,5 +1,10 @@
 ranks = %i[
-  email_rank location_rank organization_name_rank phone_number_rank website_rank
+  email_rank
+  location_rank
+  organization_name_rank
+  organization_type_rank
```
anandaroop: Not so obvious from the reformatted diff, but `organization_type_rank` was the one that was missing.

jonallured:
This is awesome, I actually ran into the missing Sentry var yesterday, so thanks for fixing that! The only thing I'm not sure about is the change to Sidekiq concurrency. What's in your PR is actually were we started, but I found it hard to mess with later. I ended up liking the declarative approach better. But that probably does mean that there should be an entry for that in the `.env.example` file so that when a new dev starts things up, they'll have a value for that one. What do you think about that?

db/seeds.rb
jonallured: Oops, thanks!!

jonallured:
Thanks for making the process better for the next dev! 💯

PR #294, opened 2017-08-15
@jonallured: fix charlock_holmes gem setup

cavvia:
The gem `charlock_holmes` depends on ICU, and [it recommends](https://github.com/brianmario/charlock_holmes#homebrew) installing that via homebrew. You then have to point bundler to the correct directory as its a keg-only install. Lastly, the old release (0.7.3) doesn't seem to compile with the latest ICU, so I updated it and now `bundle install` works fine.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -78,7 +78,7 @@ GEM
       carrierwave
       fog
       uuidtools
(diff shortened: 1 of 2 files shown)


jonallured:
Nice one!!

PR #295, opened 2017-08-17
@jonallured: add an elasticsearch index for orgs #migration

cavvia:
This adds elasticsearch indexing for Organization objects, using a `Searchable` concern that backs onto our own [estella gem](http://www.github.com/artsy/estella). A `searchable` block is placed in the model which details which fields to index.

This adds indexing callbacks on creation, updates (via `#save!` only) and destruction of Organization objects. Indexing is performed in the background using Sidekiq jobs - indexing a document should take no longer than 10ms or so.

Additionally a rake task is provided to reindex all the Organization objects.

Elasticsearch is now required to be running in order to run some of the test suites. I've added this to the `setup` script and I've also attempted adding a docker image to the circleci config, though I will probably need to play around with that to get it working. We don't quite mirror this setup elsewhere, as [hokusai](http://www.github.com/artsy/hokusai) handles all the container config for CI in our other apps using ES containers.

The next step will be to produce a search endpoint with a custom ES query for auto-completing organization names, as required for our reference show creation lookup in Volt.

**Issues**

Searching by website url currently doesn't work very well, due to our default analysers not really being designed for url strings. Estella declares four analysers for a string field defined with `FULLTEXT_ANALYSIS` and here's what the tokens look like for `http://www.example.com`: 

```
default:

[ "http", "www.example.com" ]

ngram:

[ "ht", "htt", "http", "ww", "www", "www.", "www.e", "www.ex", "www.exa", "www.exam", "www.examp", "www.exampl", "www.example", "www.example.", "www.example.c", "www.example.co", "www.example.com" ]

shingle:

[ "http", "http www.example.com", "www.example.com" ]

snowball:

[ "http", "www.example.com" ]
```

The easiest fix for this is to parse out the hostname and just index that rather than the full url. That's easier than adding more custom analysers, but it does mean the website is not necessarily linkable when presenting search results. I don't believe searching by website is our main use-case but it would be nice to have. 

Our main use-case for reference shows is auto-completion based on Organization name, and the `ngram` analyzer provides us what we need for that.

**Migration**

In a Rails console:

```
Organization.create_index!
```

In production CLI:

```
```


app/models/organization.rb
```diff
@@ -16,6 +16,43 @@ class Organization < ApplicationRecord
...
+  def cities
+    locations.pluck(:city)
+  end
+
+  def countries
+    locations.pluck(:country)
+  end
+
+  def search_boost
+    locations.size
```
cavvia: This was just the only boost that came to mind at the time. it's only applied when using the default `#es_search` call. For our API endpoint we will write a custom ES query in any case and can stipulate any boost we like there.

app/models/organization.rb
```diff
@@ -16,6 +16,43 @@ class Organization < ApplicationRecord
   validates :in_business, inclusion: { in: [CLOSED, OPEN, UNKNOWN] }
 
   include Auditable
+  include Searchable
+
+  searchable do
+    field :id, type: 'integer'
+    field :name, using: :names, type: 'string', analysis: FULLTEXT_ANALYSIS, factor: 1.0
```
cavvia: The `factor` is a relative field weighting applied only when using the default search lookup via the `#estella_search` interface.

jonallured:
I've been reading through this and it's looking really good, thank you @cavvia!!

Looks like the failures are because you've shadowed a relation name, this fixed it for me:

```ruby
class Organization < ApplicationRecord
  # ...
    searchable do
      # ...
      field :tag, using: :tag_names, type: 'string', analysis: FULLTEXT_ANALYSIS, factor: 0.5
  # ...
  def tag_names
  # ...
end
```

That way there isn't a collision between the search method and the tags relation from ActiveRecord.

cavvia:
Thanks for catching that! Updated now with the fix. I've also added the `ELASTICSEARCH_URL` config var to Heroku for staging and production.

bin/setup
```diff
@@ -28,6 +28,13 @@ if ! command -v psql > /dev/null; then
   exit -1
 fi
 
+if ! command -v elasticsearch > /dev/null; then
+  printf "${CLEAR_LINE}${RED}  You must install elasticsearch before setup can continue${NO_COLOR}\n"
+  printf "    You should:\n"
+  printf "      $ brew install elasticsearch\n"
```
jonallured: Is there a certain version of ES that should be installed. I've got 2.4 installed but I think latest is 5, correct? I assume brew would give you the latest version by default but it's been a while since I was messing around with this.

config/initializers/elasticsearch.rb
```diff
@@ -0,0 +1,11 @@
+require 'typhoeus/adapters/faraday'
```
jonallured: I'm confused about the relationships between Elasticsearch, Estella, Typhoeus and Faraday. Why did we need to add Typhoeus to the Gemfile just to use the Faraday adapter?

spec/models/organization_spec.rb
```diff
@@ -95,5 +95,32 @@
         expect(sources).to match_array([source1, source2, source3])
       end
     end
+    context 'search indexing' do
+      before do
+        Organization.recreate_index!
+        @org = Fabricate :organization
```
jonallured: Do you mind pulling this out to a `let`? #minor

config/initializers/elasticsearch.rb
cavvia: Hmm, good question! This allows the Elasticsearch API to use keep-alive connections and non-blocking requests via [Typhoeus](https://github.com/typhoeus/typhoeus) configured with Faraday as its adapter, as recommended in the [elasticsearch-transport doc](https://github.com/elastic/elasticsearch-ruby/tree/master/elasticsearch-transport). The transport lib is included by the `elasticsearch-model` gem, which is the official Ruby client used by Estella. From the doc:

> For optimal performance, use a HTTP library which supports persistent ("keep-alive") connections, such as Typhoeus. Just require the library (require 'typhoeus'; require 'typhoeus/adapters/faraday') in your code, and it will be automatically used; currently these libraries will be automatically detected and used: Patron, HTTPClient and Net::HTTP::Persistent.

We can check the transport, which shows that Typhoeus is being used as the connection handler, so it's not just the Faraday adapter that is being detected and used:

```ruby
Elasticsearch::Model.client.transport
=> #<Elasticsearch::Transport::Transport::HTTP::Faraday:0x007f8ff3c29270
Elasticsearch::Model.client.transport.connections.first
=> #<Elasticsearch::Transport::Transport::Connections::Connection:0x007f8ff3c3b5d8
 @connection=
  #<Faraday::Connection:0x007f8ff3c28fa0
   @builder=#<Faraday::RackBuilder:0x007f8ff3c28c08 @handlers=[Faraday::Adapter::Typhoeus]>,
  ...
```

Typhoeus gives us non-blocking requests, so I think it's probably worth using for that fact alone.

bin/setup
cavvia: Good point! Ideally we want to test and develop on 2.x as we use 2.1 in production - 2.1 used to be the default version via Homebrew, but it appears to be 5 now. 

Judging by `brew search`, Homebrew only offers 2.4 as a managed version currently, so maybe `brew install elasticsearch@2.4` would be better here.

spec/models/organization_spec.rb
cavvia: sure! done.

bin/setup
jonallured: Cool thats what I thought but I wasn't sure - thanks for clarifying! 🥇

config/initializers/elasticsearch.rb
jonallured: Yeah, seems legit - thanks for the documentation links!

jonallured:
Well, I think this is ready to merge and deploy - you good with that @cavvia?

cavvia:
Sure - go for it! I will then try to run the migrations on staging before prepping a deploy PR as per the Deployment doc.

Issue #296, opened 2017-08-29
Broken reindex task


jonallured:
After merging #295 and deploying to Heroku, I ran the migration steps:

```
Organization.create_index!
bundle exec rake elasticsearch:reindex
```

The first step was easy, but the rake tasked failed. The first fix was to require the environment so that `Organization` wouldn't be an uninitialized constant:

```ruby
task reindex: :environment do
```

The next error was similar - `SearchIndexWorker` should have been `SearchIndexJob`:

```ruby
Sidekiq::Client.push_bulk('class' => SearchIndexJob, 'args' => args)
```

But then I was still getting errors, so I actually put that class in as a String instead:

```ruby
Sidekiq::Client.push_bulk('class' => 'SearchIndexJob', 'args' => args)
```

That got the task to execute correctly, but then the worker logs blew up with errors:

```
2017-08-29T20:23:58.681487+00:00 app[worker.1]: /app/vendor/bundle/ruby/2.4.0/gems/sidekiq-5.0.4/lib/sidekiq/util.rb:25:in `block in safe_thread'
2017-08-29T20:23:58.683364+00:00 app[worker.1]: E, [2017-08-29T20:23:58.683309 #4] ERROR -- : Not sending event due to previous failure(s).
2017-08-29T20:23:58.683413+00:00 app[worker.1]: E, [2017-08-29T20:23:58.683366 #4] ERROR -- : Failed to submit event: undefined method `jid=' for #<SearchIndexJob:0x007f44a83225f0>
2017-08-29T20:23:58.683575+00:00 app[worker.1]: 4 TID-osu4suwhc WARN: {"context":"Job raised exception","job":{"class":"SearchIndexJob","args":["Organization",147164],"retry":true,"queue":"default","jid":"1ec3d0d13d5329e935be013d","created_at":1504038177.2648933,"enqueued_at":1504038177.2888005,"error_message":"undefined method `jid=' for #<SearchIndexJob:0x007f44901ee368>","error_class":"NoMethodError","failed_at":1504038238.6780915,"retry_count":0},"jobstr":"{\"class\":\"SearchIndexJob\",\"args\":[\"Organization\",147164],\"retry\":true,\"queue\":\"default\",\"jid\":\"1ec3d0d13d5329e935be013d\",\"created_at\":1504038177.2648933,\"enqueued_at\":1504038177.2888005}"}
[...]

PR #297, opened 2017-08-30
@jonallured: fix elasticsearch reindex task

cavvia:
Thanks for looking into the migration issues with the `elasticsearch:reindex` task. I made the two fixes you suggested and switched the workers away from `ActiveJob` to just use the `Sidekiq::Worker` module as we do in our other projects. The migration was working locally with this new setup.

ActiveJob doesn't support bulk pushing, as confirmed by Mike Perham [here](https://codexample.org/questions/721381/bulk-insert-using-activejob-and-sidekiq.c). We do want that ability in this context as it's more efficient when iterating over all the Organizations - it reduces Redis round-trip latency. 

Fixes https://github.com/artsy/bearden/issues/296.

### DIFF
--- a/app/jobs/search_delete_job.rb
+++ b/app/jobs/search_delete_job.rb
@@ -1,4 +1,6 @@
-class SearchDeleteJob < ApplicationJob
+class SearchDeleteJob
+  include Sidekiq::Worker
+
   def perform(klass, id)
     obj = klass.constantize.find(id)
     return unless obj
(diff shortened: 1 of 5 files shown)


jonallured:
Cool, I'm going to push this to staging and make sure all is well there before merging to master. Thanks for jumping on this today! ❤️

jonallured:
Well staging is done, so I'm going to merge this and I think we're close to deploying to production (I just want to get a bundle update in too).

PR #298, opened 2017-08-30
Bundle update

jonallured:
This was mostly routine, but there were two notable updates:

* Amazon released version 3 of their Ruby SDK and have moved to a modular approach. When I first did my bundle update, there were TONS of gems in the diff. I remembered reading something about this, so I switched over to requiring just the S3 gem and all was well with the world.

* The micromachine gem had a major version bump because they changed the arity of the callbacks to accept a payload. I got a bunch of failing tests on this, but there was an obvious release note about it on the repo, so I just updated the couple callbacks to throw away another argument and was on my way.

Other than that, I did update bundler too.


jonallured:
Doing a selfie-merge on this one...

Issue #300, opened 2017-08-30
Elasticsearch environments


jonallured:
@cavvia - I wasn't sure what to do with this, but I figured a GitHub issue was as good a place as any to start - let me know if you prefer to address this some other way!

After #299 deployed out to production, I went back to #295 to run the migration steps and the first thing I did was run this:

```ruby
Organization.create_index!
```

But I got an exception about the index already existing, so then I thought - oh no, maybe staging and production are pointed at the same ES! Sure enough, when I compared the `ELASTICSEARCH_URL` values on staging and production I discovered they are the same. So, this might be intentional, or not - I'm just not sure and it caught me by surprise, so I wanted to bring it up. I have not run the rake task to reindex on production.

My expectation would be for there to be two separate ES environments so that what we do on staging could not affect production - is that how Artsy typically uses ES? Thanks!

cavvia:
We do use the same `ELASTICSEARCH_URL` in staging/production - we use the same cluster, but usually create separate indices within that cluster. 

So we should have `organizations_staging` and `organizations_production` indices within that cluster. The issue is that I used `RAILS_ENV` to determine the index name, and we actually run staging on `RAILS_ENV=production` as Rails does not have a staging environment. So the migration you ran in staging has populated the `organizations_production` index. You can check this in a console, using `httpie` like so:

```shell
http $ELASTICSEARCH_URL/_cat/indices
...
green open organizations_production        1 1  109515      0 148.6mb  75.5mb
```

That shows a fully populated production index with 109k documents.

We encountered this issue in Diffusion too - in that case I just created a new env var for the index name, which we could set separately in staging and production. I can do that here too and then we can run both migrations again.

jonallured:
Ah ha! Makes total sense and it sounds like the approach you took on Diffusion would work well for Bearden. Thanks for the info!! 🤗

PR #301, opened 2017-08-31
@jonallured: fix elasticsearch environments

cavvia:
As discussed in https://github.com/artsy/bearden/issues/300, I use a new ENV var called `ELASTICSEARCH_ENV` to deduce the index name for Organizations.

Thanks for catching that issue - it recently caught me out in Diffusion too, so I really should have foreseen it.

I've updated the config vars in Heroku staging/production.

**Migration**

We need to re-migrate now I guess, since all our staging data incorrectly went into a production index. So on staging first, and then production, we should do this:

```ruby
# this recreates the index and all its settings/mappings without reindexing the data.
Organization.reload_index! 
# exit the rails console...
```
 
I'm happy to do that myself once this is merged, but let me know if you'd rather do it. 

Fixes https://github.com/artsy/bearden/issues/300

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -5,7 +5,7 @@ module Searchable
   include Estella::Searchable
 
   included do
-    index_name [name.downcase.pluralize, Rails.env].join('_')
+    index_name [name.downcase.pluralize, Rails.application.secrets.elasticsearch_env].join('_')
 
     # disable estella inline indexing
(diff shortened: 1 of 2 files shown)


jonallured:
Nice one - I'll babysit this one out to staging and production. 💯

jonallured:
Ok, this is merged, deployed and migrated on both staging and production - thanks so much for your help @cavvia!! ❤️

PR #303, opened 2017-09-05
@jonallured: Organizations Search API. #migration

cavvia:
This adds an Organization Search API using [Grape](https://github.com/ruby-grape/grape), mounted at `/api/v1/organizations/search`, which takes a `term` query param and supports pagination. The search backs onto Elasticsearch and is tailored towards auto-complete, as I believe that's the first client use-case presented by the Artist CV project.

For authentication I use JWT tokens as we do elsewhere (Impulse, Radiation, etc). This is our standard for microservice auth. We expect the `aud` (audience) property to contain the name of a trusted app, encoded using a Bearden secret key. There is no user authorization or roles at this stage - a trusted app simply has full access to the API. This seems sufficient for the search API as it seems fine to me that any user of Volt should have access to it. I have added `volt` and `bearden` as trusted apps for now.

I have updated the README with some details on how to consume the API using JWT.

**Migration** 

A JWT secret has to be placed in the staging and production `JWT_SECRET` ENV var. We have been using standard 128-bit hex strings - that's not the most secure so I would double the size to use a 64 character key instead, via SecureRandom. For example: 

```ruby
require 'securerandom'
SecureRandom.hex(32)
=> "38bb35000902bf28a4b7c34df33e3f26839dffec4cfdd2a859a5ba6fb7acfac5"
```

I will generate a unique secret key for each env in this manner and set them in Heroku.

This contributes towards https://github.com/artsy/collector-experience/issues/354.

cc @ashkan18 who has experience using JWT - if you could just verify I am using JWT in the correct manner here, as this is my first time using it. I have largely lifted the middleware and helpers from Impulse so it should all look familiar.


cavvia:
Migration has been applied.

app/api/organizations/api.rb
```diff
@@ -0,0 +1,21 @@
...
+  class API < Grape::API
+    version 'v1', using: :path
+    format :json
+
+    include Authentication::Helpers
+
+    namespace :organizations do
+      desc 'auto-complete organization names'
+      params do
+        requires :term, type: String, desc: 'Term to match.'
```
ashkan18: are we indexing `lat`, `lon` and locations too? i was wondering if it makes sense to add those to our query too. That can totally be separate PR if indexing by location is needed.

README.md
```diff
@@ -51,6 +51,18 @@ PRs merged to the `master` branch are automatically deployed to staging.
 Production is automatically deployed upon merges to `release`. Create such a PR
 with [deploy_pr][deploy_pr] or [this handy link][deploy].
 
+## API
+
+Trusted apps may use the Bearden API using [JWT](https://jwt.io/) authentication. You will need to obtain the JWT secret for Bearden, please ask a Bearden developer for this. Once in possession of the secret, you need to encode a JWT token using the [JWT gem](https://github.com/jwt/ruby-jwt):
```
ashkan18: I would suggest to use JWT created by Gravity in this case too. Thats what we do in Impulse as well, in that scenario `JWT_SECRET` is basically `internal_secret` of that `ClientApplication` in Gravity. In order to create a new JWT, then we can refer to [this](https://github.com/artsy/gravity/blob/master/doc/ApiAuthentication.md#create-xapp-token) doc.

cavvia:
@ashkan18 Makes sense to use the gravity `internal_secret` for the Bearden ClientApplication instead of creating a new one. I can update to use that. I'm still not fully clear how clients request Gravity for the JWT however. The doc you link to refers only to an XAPP token. How does that relate to the JWT the client needs to send to Bearden?

app/api/organizations/api.rb
cavvia: We only index city and country names as I don't believe we have an actual `lat` and `long` for these Organizations, maybe @jonallured can clarify if we do. We could add filters for country and city very easily to this request, as additional params which we send to ES. I'm not sure there's any need for that right now, but definitely an enhancement.

ashkan18:
if it's a trusted client, we need to create a JWT using that command I posted, basically we find `Bearden` client application and we create a `ApplicationTrust`.

If user's want to access Bearden, they need to first call Gravity's @POST /api/v1/me/token` and ask for Bearden JWT.

cavvia:
Your scheme also fixes my abuse of the `aud` (audience) field, which is intended to be "the recipient of the token" - I was mistakenly using it to identify the sender.

jonallured:
I hate to chime in late and after so much work has been done, but I have to ask about Grape. Can you help me understand why Grape is a good fit for adding this endpoint?

Another question I have is why such an old version of the gem - I see that 0.17.0 was picked and that version was released over a year ago. I also see that version v1 was released about a month ago - is there some reason you didn't want to use that version?

Sorry @cavvia, we probably should have touched base on these points before you got started, but I didn't know to ask these questions earlier!!

cavvia:
@ashkan18 I've updated this to use a Gravity JWT. Could you check the README to see if it makes sense?

@jonallured I used Grape as it's a fully featured API framework we use in a number of services (Gravity, Impulse, etc). It's mainly geared towards REST APIs, but it basically had the features I needed for mounting endpoints, handling params and adding auth. What would be your recommended approach? Clearly I don't want to mix these endpoints up with the web application. The version is old because I cargo-culted the Gemfile entry from Gravity, I can definitely update it.

jonallured:
Just wanted to drop in and update this PR - there was a little discussion on Slack about this topic yesterday. I clarified that I was not suggesting GraphQL, but rather, I'd like to see the Bearden API start its life as simple Rails controllers. I feel that adding Grape goes against my goal of keeping the project a pretty straightforward Rails app.

I'm about to have some time off and this work is not on deadline, so I'll come back to this next week.

app/api/authentication/helpers.rb
```diff
@@ -0,0 +1,34 @@
+module Authentication
+  module Helpers
+    extend ActiveSupport::Concern
+
+    included do
+      use Middleware::JwtMiddleware
+      rescue_from JWT::DecodeError do |_e|
```
damassi: Does the `_e` convention mean "unused argument `e`"?

app/api/authentication/helpers.rb
```diff
@@ -0,0 +1,34 @@
...
+          return if current_app == Rails.application.secrets.artsy_application_id
+          error!('Unauthorized', 401)
+        end
+
+        def jwt_payload
+          @jwt_payload ||= env['JWT_PAYLOAD']
+        end
+
+        def current_app
+          @current_app ||= jwt_payload&.fetch('aud', nil)
```
damassi: Is the `&` an existential operator that guards against a missing `fetch`?

app/api/authentication/helpers.rb
jonallured: Yeah, totally - sometimes you even just do this:

```ruby
rescue_from JWT::DecodeError do |_|
```

app/api/authentication/helpers.rb
jonallured: Yeah, they call it the Safe Navigation Operator:

http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

cavvia: That's right - it guards against a nil `jwt_payload` in this case.

app/api/authentication/helpers.rb
cavvia: I would usually just use `_` as @jonallured points out. That's also convention in some other languages (eg Scala).

jonallured:
Had a great pairing session with @cavvia yesterday where I showed him how I have done simple APIs with Rails in the past. We worked on a ping endpoint and then started on the search endpoint, but ran out of time before finishing up. We agreed that I'd merge this PR and then I'd do another PR on top that replaces Grape with a Rails version.

spec/models/organization_spec.rb
```diff
@@ -95,32 +95,34 @@
...
+      org.locations.create! content: 'Berlin, Germany', city: 'Berlin', country: 'Germany'
+      tag = Fabricate :tag, name: 'fooy'
+      ot = Fabricate :organization_tag, organization: org, tag: tag
+      org.organization_tags << ot
+      org.websites.create! content: 'http://www.example.com'
+      org.organization_names.create! content: 'Quux Gallery'
+      org.save!
+      Organization.refresh_index!
+      # using the default estella query
+      expect(Organization).to receive(:estella_search_query).at_least(:once).and_return(Estella::Query)
```
jonallured: Hey, quick question @cavvia - why was it helpful to swap out the `OrganizationsQuery` for the default one?

PR #305, opened 2017-09-06
Fix broken nav

jonallured:
When I upgraded Bootstrap all the tests passed, but the navigation was broken and I didn't ever actually spin up a server and check. :cry:

This commit fixes that by adjusting a couple classes. I also fixed a mistake where we weren't highlighting syncs when that was the page you were on.


jonallured:
Doing a selfie-merge on this one so I can get it fixed on production ASAP.

PR #307, opened 2017-09-06
Bundle update

jonallured:
I removed some version pins that I wasn't sure mattered. The tests are green and I can't seem to find any reason to pin them, so I updated the latest Typhoeus.


greysteil:
Hey @jonallured - think I might have asked before, but would you be up for giving [Dependabot](https://dependabot.com) a try on Bearden to automate dependency update PRs? It's totally free - I'm just trying to get as many people using it as possible so the data we share on [compatibility scores](https://dependabot.com/compatibility-score.html) is as good as possible.

PR #308, opened 2017-09-15
Use Rails for API

jonallured:
This PR started as a pairing session with @cavvia where we worked together to recreate the API work he had done on #303 in straight Rails. I completed this work and have pulled out grape as well.

The first step was to extract an AdminController so that it would be much harder for the API side of things to mingle with the admin side of things. There's a decent amount of diff noise just from that, but all it really boils down to is ripping everything out of ApplicationController except the SSL redirects.

From there we added a Ping endpoint for no other reason than I like having them in APIs that I work on! 😂 

Once that was dealt with, we switched gears and started on the search endpoint. This was about as far as we got, so I spent some time this afternoon and finished it up. One goal I had was to keep the API-level specs focused on request/response stuff and move the specs around search results down to the model-level. That's why you'll see some diff around the Organization specs. That feels like a better home for these assertions.

Then it was time to remove grape - I did it in one commit by just removing the gem and them rm -rf a couple folders and seeing what broke. There's a lot of noise in the diff from this too. I then reviewed the diff from #303 to catch anything that I might have missed. Pretty sure I got it all, but please let me know if I missed something!

Finally, I did pull up the logic around ensuring the JWT token is passed in correctly to the ApiController - this might be a little premature, but we have an endpoint in mind that adds organizations, so it felt appropriate to me.


app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Api::V1::SearchController < ApiController
+  before_action :ensure_term
+
+  expose(:organizations) { Organization.estella_search(estella_options) }
+
+  private
+
+  def estella_options
+    params.permit(:term)
```
cavvia: This should also accept `size` and `page`  params for pagination. The latter is marshalled into a `from` offset as that's what ES works with (see [here](https://github.com/artsy/bearden/pull/308/files#diff-8b9642288245a75a3e8b37781a17c935L16)). I use `page` instead of `from` as that's the convention for pagination in all our core APIs.

cavvia: We can also add a spec for pagination params but it's pretty simple stuff.

app/views/api/v1/search/index.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.array! organizations, :names, :tag_names, :website_urls, :cities, :countries, :id, :in_business
```
cavvia: Interesting that the json builder is now API specific whereas previously the `as_json` was just a generic json representation declared by the model itself. Will each API have to decide how it wants to represent organizations? While this gives flexibility, it's nice to have a default representation as that will standardize our API responses and it means the API doesn't have to decide each time what an Organization is.

cavvia:
Looks good! We just need to add pagination support back in I think.

It definitely makes sense to have search specs at the model level too. We do this elsewhere, I think I had cut a corner there just including API integration specs 😛 .

app/controllers/api_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def token_match?
+    auth_header = request.headers['HTTP_AUTHORIZATION']
+    return unless auth_header
+    token = auth_header.split(' ').last
+    secret = Rails.application.secrets.artsy_internal_secret
+    JWT.decode(token, secret)
+  rescue JWT::DecodeError
+    return false
+  end
```
cavvia: :+1: I agree, this should definitely live here.

app/controllers/api_controller.rb
```diff
@@ -0,0 +1,20 @@
+class ApiController < ApplicationController
+  protect_from_forgery with: :null_session
```
cavvia: Since auth is JWT based and there is currently no user auth for the API, I wonder if CSRF and related methods affect us at all?

spec/models/organization_spec.rb
```diff
@@ -96,33 +90,103 @@
...
+        tag: Fabricate(:tag, name: 'fooy')
+      )
+
+      Fabricate(
+        :website,
+        organization: organization,
+        content: 'http://www.example.com'
+      )
+
+      organization.es_index
```
cavvia: I don't think an explicit call is required, as it should automatically index on creation via callbacks.

cavvia: I just noticed the object is created prior to the `Organization.recreate_index!` call, so I see why you've included this now!

app/controllers/api/v1/ping_controller.rb
```diff
@@ -0,0 +1,7 @@
+class Api::V1::PingController < ApiController
+  skip_before_action :ensure_jwt
+
+  def show
+    render json: { timestamp: Time.now.to_i }
```
cavvia: Good idea to have a ping endpoint as we can tie our pingdom alert to this once the API is in use by Volt.

app/views/api/v1/search/index.json.jbuilder
jonallured: Yeah, I see what you mean - it's possible that duplication will creep in here, so something to watch for, but I like this info being in the view layer. In my experience you either quickly discover something to extract or it turns out the duplication is fine.

app/controllers/api_controller.rb
jonallured: I think that will this in place CSRF should not apply to these endpoints, which is what I was going for. Without an actual client, I don't know for sure - happy to address this once engineering hits this part of the process.

spec/models/organization_spec.rb
jonallured: ❤️

jonallured:
@cavvia: awesome, glad you liked how this turned out - I'm happy with it too!!

Regarding pagination - I was thinking about this PR over the weekend and it occured to me that I didn't address this point so I'm glad  you brought it up! I touched base with @madeleineb on this and confirmed with her that the only use case for this endpoint is a gallery name autocomplete, so we don't actually need pagination at this point! There are ideas about other consumers of this endpoint in the future that might want to page through, but I'd prefer to punt on this until we have a solid requirement.

So, question for you: how would you go about hard-coding a results size of 20? If I remember correctly, I think you said that by default ES will return 1,000 results, so what I'm trying to figure out is how to tell it that I only want 20 - would that be something like this:

```ruby
Organization.estella_search(term: term, page: 1, size: 20)
```

I checked out that line you linked to above:

```ruby
params[:from] = (params[:page] - 1) * params[:size]
```

But I was a little confused about how `from` interacts with `page` and `size`.

Thanks!! ❤️ ❤️ ❤️

cavvia:
Makes sense with respect to pagination - you can hardcode the results size just using the `size` param in the `estella_search` call. It would be nice (but not a blocker) if the client could send this as a query param since otherwise a change in the client UI would necessitate a code change in Bearden.

`#estella_search` does not recognize the `:page` param, only `:from` which is basically an offset. That's consistent with the ES API. So I was deriving the `:offset` from the page/size API query params, as these are our conventional API pagination params in our Core API (Gravity).

jonallured:
Cool, then what I'll do is allow the size to come from params and just default it to 20. That way the client can choose what they'd like to do. And thanks for the explanation - I get it now. 👍

spec/requests/api/v1/search_spec.rb
```diff
@@ -63,6 +63,44 @@
...
+        end
+
+        Organization.refresh_index!
+
+        size_override = default_size - 1
+        get '/v1/search', params: { term: 'Gallery', size: size_override }, headers: headers, as: :json
+
+        response_json = JSON.parse(response.body)
+        expect(response_json.count).to eq size_override
+      end
```
cavvia: :+1: very clear specs here

cavvia:
Looks great now with the `size` param. Going to merge this in. :clap:

dblock:
I'm late to this discussion, but there's a similar thread in https://github.com/artsy/constellation/issues/35 around what the API should look like. Seems like we're going in many different directions across GraphQL, HAL, now Rails API (what is the data layout here?). What was the motivation for this?

jonallured:
Hi @dblock, I'd be happy to have a more in-depth discussion, but the tl;dr is that I felt Grape was too much for just a single endpoint. @cavvia and I went back and forth on it a little and then he was gracious enough to setup a time to pair on this so that I could show him what I was thinking. My motivation was and has been to keep this project as boring of a Rails app as possible. 😴

dblock:
Thanks @jonallured. I don't really care about Grape vs. Rails, more about API data layout and versioning. Basically why isn't this a hypermedia API and why does it live under `/v1`?

jonallured:
The only endpoint we have here is a search endpoint that takes a term and spits back some data about matching organizations. We don't even have a client for this endpoint yet, so technically, this is probably a little premature, but @cavvia was interested in working on this project during pair swap, so he took it on. The reason I went with `/v1` was because I always do that. What do you think we should have done differently?

dblock:
Make this like the Gravity API v2.

1) Kill `/v1` and make versioning use Accept headers. In Gravity we never added v2, the v2 API is just `/api`. [This](http://code.dblock.org/2013/07/19/evolving-apis-using-grape-api-versioning.html) could be helpful.
2) Use [roar-rails](https://github.com/apotonick/roar-rails) and turn this into a hypermedia API in HAL format with a root. This way you can use a generic client like [hyperclient](https://github.com/codegram/hyperclient) and expose organizations as first class objects. This way your search endpoint returns objects of the same type that are individually accessible (ie. you only have 1 way to retrieve data, right now you return organizations in search and you don't have a way to retrieve one organization).

jonallured:
Well, those could be great patterns to follow if/when this project gets there, but I think jumping to them right now is premature optimization. What's right for Gravity is not necessarily right for other projects.

dblock:
I respectfully disagree with this statement - these aren't optimizations, these are questions of API design.

PR #310, opened 2017-09-22
@jonallured => Reimplement a HAL API.

dblock:
This rewrites the API from #308 in Hypermedia format and gets rid of `/v1`. 

The differences are subtle, really just adds a root endpoint and forces a certain style of data (HAL), but very important for a naiscent API. Why is this better?

1) the API is hypermedia, self-documenting and clients follow links
2) standard HAL format means you can use a generic client, much nicer for client developers
3) standard HAL means there's a standard way to add fields and link classes of things together with links, including links to objects outside of the Bearden application if needed
3) standard HAL means we have an easy way to add, for example, a count of objects or a cursor-paginated collection, and not have to hack returning some X-Total-Count header or other hack
4) representers are much cleaner than json views and can have logic
5) tests are much easier to write and a lot less code that doesn't need to know it's HTTP/JSON, never need to `JSON.parse`
6) naming of endpoints is consistent and future proof, only nouns, not verbs (organizations vs. search)
7) versioning with accept headers can come later as needed without having to fork the entire api

Some optional Hypermedia reading:

https://medium.com/unexpected-token/how-your-api-could-benefit-from-hypermedia-b62780771ccb


orta:
Ah, this reminds me that I wanted to run a meeting with everyone in dev about the standardizing of this stuff. Will see if there's a good time for that next week.

jonallured:
I spent some time with this PR this morning and wanted to get my thoughts out in a comment before they left my brain. I started by making a list of all the things this PR does:

* adds a RuboCop todo file
* adds four gems
* moves the API documentation from the README to the docs folder
* moves the ping endpoint to a status endpoint
* moves the search endpoint to an organizations endpoint
* adds the show endpoint for organizations
* adds a root endpoint
* switches from jbuilder view templates to representers
* moves request specs to "api" specs
* adds new mime type

But the biggest change is what you talk about in your PR description - moving the API from something pretty loose to something pretty standard. As I've said before, I'm not convinced, but I don't feel super strongly about this either and it seems like you really do, so I'm happy to move in this direction. Especially since you already did all the work! 😉 

My point of view on APIs comes from my experience where I have control of both sides (client and server) and everything is internal. Which is actually this situation as well - Volt would be the only client.

My approach to software development is to do just the minimum, so my plan was to start with as simple an API layer as possible and then build whatever was required once the API started getting used. I didn't feel what we had was done, only that it was ready for the client to start consuming. Once we had real-world usage, we'd let that feedback drive out whatever else needed to be one.

I do have a couple unrelated questions:

* what's up with RuboCop - why add the todo file?
* why add the new show organization endpoint?

And if anyone else has any thoughts, I'd be very interested to hear them as  well! ❤️

mzikherman:
I have some general thoughts re: HAL and my experience (and was considering it for Constellation), since there's a nice writeup of some pros in this PR, I thought I'd respond to them. The tl;dr is I'm still unconvinced/confused by HAL at times, and have generally been successful with 'custom' Rails/JSON and GraphQL.

> the API is hypermedia, self-documenting and clients follow links

This is great for a public API, but doesn't have a huge benefit when the primary consumer is in-house clients or intra-app messaging.

> standard HAL format means you can use a generic client, much nicer for client developers

GraphQL, and more targeted REST queries are pretty simple HTTP requests too, personally I've been tripped up using `Hyperclient` in the past, maybe I'm overthinking it.

> standard HAL means there's a standard way to add fields and link classes of things together with links, including links to objects outside of the Bearden application if needed

> standard HAL means we have an easy way to add, for example, a count of objects or a cursor-paginated collection, and not have to hack returning some X-Total-Count header or other hack

I would call returning `X-Total-Count` as 'standard' as well, standard Artsy that is! Seriously though, the general pros here for more widely adopted standards are great for public APIs, but are less so for the dominant use case of in-house clients/APIs.

> representers are much cleaner than json views and can have logic

I think the same thing can be accomplished in GraphQL resolvers, JSON views/builders, Grape endpoints, etc.

> tests are much easier to write and a lot less code that doesn't need to know it's HTTP/JSON, never need to JSON.parse

I don't know if this is a huge pro, I think request/integration specs would all be pretty similar.

> naming of endpoints is consistent and future proof, only nouns, not verbs (organizations vs. search)

I guess, my feelings are similar to the public vs in-house issue.

> versioning with accept headers can come later as needed without having to fork the entire api

This I don't know much about at all, but I feel that it's more about public vs in-house stuff, but perhaps I don't have enough experience :)


I hope I haven't revealed myself to be closed-minded, or not 'getting it', or something else- but my personal 2c is pretty much in line with @jonallured- vanilla Rails controllers/JSON, esp in the beginning of the project when expectations are unclear. If the project never takes off, or simply settles into that early work being sufficient- cool, it can probably just remain as-is. If more complicated front-end or other needs require a different/revamped API, that can then be added (maybe GraphQL in general, or perhaps Grape/JSON for 'custom' endpoints, if Rails/CRUD feels awkward).

dblock:
@jonallured 

> what's up with RuboCop - why add the todo file?

I typically just run `rubocop -a ; rubocop --auto-gen-config` and don't worry about it. This helps enforce ruby coding standards that rubocop provides and not think about it, the latter generates a todo for exceptions.
 
> why add the new show organization endpoint?

I removed "search" in favor of "organizations", because "search" is not RESTful, but smells like an RPC method that may or may not return organizations

dblock:
@mzikherman 

You're saying "leaving things as is" because this is an internal API? Aren't Artsy developers people too? Why should it matter than an API is internal vs. external?

dblock:
On internal vs. external: https://apievangelist.com/2012/01/12/the-secret-to-amazons-success-internal-apis/

mzikherman:
> You're saying "leaving things as is" because this is an internal API? Aren't Artsy developers people too? Why should it matter than an API is internal vs. external?

@dblock  I meant that- in the case of the RSVP back-end/front-end for instance, a simple Rails app wound up being all that was needed. There's only 1 (maybe 2) developers that ever interface with that app, there's just no more need for further development on it. So in that case, starting with HAL would have been a bit cumbersome, IMO. And it turned out the app settled into a 'stable' state where it was relatively feature complete, and adding newer features (mostly by a single dev on the back and front-end), can be done organically in the existing framework.

I guess I'm sort of thinking about different 'kinds' of internal apps. The bigger and public Gravity, as well as Impulse which serves multiple clients, vs Admin Metadata, Galaxy, RSVP, which are so tiny that I'd very much want to keep super simple.

Maybe we can settle/agree that the super focused tiny apps like those, can benefit from the Rails controller (simplest possible one), and bigger and more adopted ones can/should use HAL?

dblock:
Sorry I am confused @mzikherman why is the devbate "rails controller" vs. "hal"? This PR uses Rails controllers.

mzikherman:
Ah, sorry- I keep saying 'Rails controller' and basically meaning, JSON representation of objects/lists. Not hypermedia, links, etc.

dblock:
Just so we're saying the same thing - HAL is a registered hypermedia type, but it's still JSON representation of objects, albeit a slightly more disciplined one.

cavvia:
Matt and Jon make some good points here I think - ease of client implementation (discoverability, standardized relations) is one of HALs great advantages, and that comes into its own when there are going to be hundreds or thousands of clients consuming an API. Amazon is a good example of a company reaching the size where internal APIs should be designed with that criteria in mind. Standardizing for ease of client use presents a trade-off, as it lowers the flexibility at our disposal when creating APIs.

Amazon's teams are famously designed with autonomy in mind. They don't need to talk to each other in person, they can just consume each others APIs. We are a small tech team that should behave like one - we should play to our strengths, namely close communication between app and service developers and the flexibility that can provide. That leads me to consider being looser about our API design standards rather than stricter. 

That's not to say we shouldn't adopt principles - I think consistent representation of resources is an important API principle regardless of framework, and something your PR handles really well with the presenter abstraction. But as @mzikherman says, we can look to do that within any API framework.

HAL also does a good job of standardizing how relations are represented. But Bearden is a microservice with one focus - Organizations and their attendant metadata. So relations aren't a primary concern in the API design at this moment.

The PR shows an interesting limitation of HAL, and how it limits our flexibility - this is a search endpoint which is forced to live at an `/organizations` route (which non-intuitively now takes a `:term` param) because arbitrary actions like `search` are not discoverable in HAL (unlike in [Thrift](https://thrift.apache.org/) or [SIREN](https://github.com/kevinswiber/siren)). @dblock How would you approach exposing actions like search in HAL? I wonder if there are conventions for that.

As you can tell, I'm in favour of an ecology of API types. I am an advocate of RPC in some cases for example, probably the most common API paradigm for internal services in our industry. I'm also in favour of transport layers other than JSON+HTTP (e.g. MsgPack+TCP for high throughput). 

I think our APIs should instead be driven by a set of shared principles rather than adopt a single Hypermedia type. Hopefully our discussion today will help to clarify some of those.

dblock:
@cavvia I think search here is a great example of coding for a specific client's needs - the service exposes organizations, not search - the latter is a concern of the application using a service not the other way around

If search only returns organizations (and I believe it will here), a compromise could be to expose `search` as a link which does a 302 to `/organizations?term=`.

dblock:
I'll close this.

anandaroop:
ceilingcatwider

Yes, thanks all! I definitely benefitted from this discussion.

jonallured:
Agreed, thanks everyone for sharing your ideas!! The town hall yesterday was really good too. FYI: I reached out to @craigspaeth to see if he'd pair with me on a GraphQL version of this PR too, so hopefully that will happen and we can compare that approach as well. ❤️

dblock:
@jonallured feel free to take over #311 !

PR #311, opened 2017-09-27
Reimplement Search API endpoint in GraphQL.

dblock:
Based on discussion in #310.

Without  redesigning the search API, replaces it with GraphQL.


app/controllers/graphql_controller.rb
```diff
@@ -0,0 +1,11 @@
+class GraphqlController < ApplicationController
+  def execute
```
ashkan18: are we going to have any authentication/authorization here?

So far in other services i've been using Grape layer authentication to decode JWT and find `user_id` and `role` and pass it as `context` to resolver [exmaple](https://github.com/artsy/gravity/blob/master/app/api/graph/root_endpoint.rb#L21)

Graphql ruby also allows you to define `middleware` for each schema [example](https://github.com/artsy/lewitt-api/blob/master/app/api/graph/schema.rb#L5)

app/graphql/types/query_type.rb
```diff
@@ -0,0 +1,12 @@
+Types::QueryType = GraphQL::ObjectType.define do
+  name 'Query'
+  description 'The query root of this schema.'
+
+  field :search, !types[Types::OrganizationType] do
+    argument :term, types.String
```
alloy: Should this be a required argument?

app/controllers/graphql_controller.rb
dblock: So I didn't implement it, but https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/13601/diffs looks interesting.

app/graphql/types/query_type.rb
cavvia: given the current implementation it would need to be yes. in other search endpoints we fall back to a `match_all` query when no term is provided, but this endpoint doesn't support that.

dblock: I made it required.

dblock:
Looks like people would like this merged. In https://github.com/artsy/bearden/pull/311/commits/e685315cb50c5b42a0df321f1f700635b045f884 I have removed APIv1, so the remaining question is auth and docs.

@ashkan18 @jonallured What should I do here? How would the graphiql interface be usable by a common developer trying to talk to this API once I stick auth on top of it?

ashkan18:
@dblock assuming we'll use JWT for this endpoint, a developer can follow [this](https://github.com/artsy/gravity/blob/master/doc/ApiAuthentication.md#create-an-app-trust-token) to get a gravity JWT for Bearden. Once we have JWT we set `Authorization` header in Graphiql to `Bearer <jwt token>` and we should be able to use GraphiQL to test the schema.

dblock:
Any idea how to fix the build failure? ES is available & al.

```
  1) GraphqlController execute runs a query
     Failure/Error: Organization.estella_search(term: args['term'], size: 20)
     
     Elasticsearch::Transport::Transport::Errors::NotFound:
       [404] {"error":"IndexMissingException[[organizations_test] missing]","status":404}
     # ./app/graphql/types/query_type.rb:9:in `block (3 levels) in <top (required)>'
     # ./app/controllers/graphql_controller.rb:6:in `execute'
     # ./spec/graphql/search_spec.rb:6:in `block (3 levels) in <top (required)>'
```

alloy:
@dblock If you haven’t yet, you may want to use the [standalone app](https://github.com/skevy/graphiql-app) version of GraphiQL.

dblock:
@alloy Is the best practice not to add `graphiql-rails` into this project in dev mode or leave it?

alloy:
I was thinking about that as well. The only reason I can think of to include GraphiQL on the server is so you can link people to a query. We haven’t been doing that recently, but that’s mainly because it broke when we added oauth infant of MP’s GraphiQL.

Maybe we could add link generation and pasteboard copying to GraphiQL.app?

orta:
I don't want to derail but there's this which can handle that: https://github.com/andev-software/graphql-ide

dblock:
This is now ready to merge. I added JWT auth, documented the API. Who's making the call?

The ping endpoint was nuked, not sure whether it makes sense or whether we want to reimplement something with GraphQL as well. Otherwise this is exactly the same function as before.

jonallured:
Cool, I will make the call and merge this - just need a little more time! 😅

app/graphql/types/query_type.rb
```diff
@@ -0,0 +1,12 @@
+Types::QueryType = GraphQL::ObjectType.define do
+  name 'Query'
+  description 'The query root of this schema.'
+
+  field :search, !types[Types::OrganizationType] do
```
cavvia: The semantics here is a little confusing - 'search' is a `field` on a `QueryType` root, but I imagine adding an endpoint to fetch Organizations by ID would be an 'organization' field in the same query root. That sounds strange, as we end up mixing entities with arbitrary operations on a flat query namespace. Is there some best practice for how to structure that better? Can they be nested for example?

I'm also wondering what the `!` signifies here - we don't use it in our declarations in GravQL (like [here](https://github.com/cavvia/gravity/blob/master/lib/graph/queries.rb#L7)).

docs/API.md
```diff
@@ -0,0 +1,34 @@
...
+expires_in = 42.years.from_now
+token = ApplicationTrust.create_for_token_authentication(app, expires_in: expires_in)
+puts token
+```
+
+Save your token in your `.env` file if needed. Once in possession of the token, send it in your requests to Bearden:
+
+```ruby
+headers = { 'Authorization' => "Bearer #{token}" }
+```
```
cavvia: this is great auth doc, very clear :+1:

cavvia:
Thanks for this - very educational. I like the type introspection and the concise implementation. I had one question about how we structure the API going forward.

app/controllers/api_controller.rb
```diff
@@ -5,12 +5,12 @@ class ApiController < ApplicationController
   private
 
   def ensure_jwt
-    render plain: 'auth is wrong', status: :unauthorized unless token_match?
+    render plain: 'Access Denied', status: :unauthorized unless token_match?
```
cavvia: I had a look and the GraphQL spec doesn't seem to include auth so I guess we can do what we like here in terms of response.

app/graphql/types/organization_type.rb
```diff
@@ -0,0 +1,11 @@
...
+  name 'Organization'
+
+  field :id, !types.ID
+  field :names, !types[types.String]
+  field :tag_names, types[types.String]
+  field :website_urls, types[types.String]
+  field :cities, types[types.String]
+  field :countries, types[types.String]
+  field :in_business, types.Boolean
+end
```
cavvia: The rigidity of this schema, which echoes our ES index schema and not our ActiveRecord schemas, might be a problem down the line, if we want to return a fuller representation of some of these entities. 

For example,  Location objects attached to an Organization, which in ES are flattened into arrays of strings but are actually first-class entities in Bearden, would need to be added in something like a `Locations` array. These would contain `country` and `city` attributes so you will get some strange duplication as this schema builds out.

Another option is to define this as an `OrganizationSearchDocument` type, but then you get divergence in how we represent Organizations in different API responses. This issues affects every API design tbh, I'm just thinking through how best to deal with it within GraphQL.

app/graphql/types/query_type.rb
dblock: The `!` means "required", but in this context I believe it does nothing, so removed.

I am with you wrt API query schema. This PR wanted to refactor the existing API "as is", so lets try to do what you suggest after it's merged? And I am a GraphQL noob so maybe @craigspaeth can help us out with what this should be?

app/graphql/types/organization_type.rb
dblock: I really think this should return organizations, but you're absolutely right about these other fields. Again all I tried to do is reimplement the existing search API, so lets do a proper exercise of modeling after this PR.

jonallured:
I finally had a chance to really read through this and I agree with @cavvia that this approach is looking great! Being a GraphQL noob, I too have some hesitation around some of the conventions around what to call things or whatever, but that can all be improved over time as we all learn together.

I have some minor things that I'll nitpick, but my only big ask is an extraction at the controller level - I'd like to see the code that constructs the query pulled out into another class. I think the `GraphqlController` is doing too much, what if it looked more like this:

```ruby
class GraphqlController < ApiController
  def execute
    result = BeardenQuery.run(params[:variables])
    render json: result
  end
end
```

That way the details of query construction aren't at the controller level and can be reused by any object in the system. Again, open to different names, but wanted to share my thought here.

.rspec
```diff
@@ -1,2 +1,3 @@
 --color
 --order random
+--format documentation
```
jonallured: I feel strongly that the progress format is the way to go, was this changed for a reason?

.rubocop_todo.yml
```diff
@@ -0,0 +1,7 @@
+# This configuration was generated by
+# `rubocop --auto-gen-config`
+# on 2017-10-02 16:26:07 -0400 using RuboCop version 0.50.0.
+# The point is for the user to remove these configuration records
+# one by one as the offenses are removed from the code base.
+# Note that changes in the inspected code, or installation of new
+# versions of RuboCop, may require this file to be generated again.
```
jonallured: I think this is a left-over and should be removed, no?

docs/API.md
```diff
@@ -0,0 +1,34 @@
+## Bearden API
```
jonallured: Thank you so much for these docs!! ❤️

app/graphql/types/query_type.rb
alloy: The `!` means non-null, you should definitely include it _everywhere_ where null will never be returned. I.e. does this return null or an empty list when no matches are found?

spec/requests/api/v1/search_spec.rb
```diff
@@ -1,106 +0,0 @@
...
-        default_size.times do
-          organization = Fabricate :organization
-          Fabricate :organization_name, organization: organization
-          organization.es_index
-        end
-
-        Organization.refresh_index!
-
-        size_override = default_size - 1
-        get '/v1/search', params: { term: 'Gallery', size: size_override }, headers: headers, as: :json
```
jonallured: I think there's a regression here around the number of organizations returned - is there a GraphQL convention for this concept that we can use?

app/graphql/types/organization_type.rb
alloy: Agreed, return a single type from anywhere where the client could expect it to be the same.

There is no need for the schema to exactly resemble the API response, if it’s logical for fields to be nested under `location` then I’d suggest you do so now. We do the same in MP.

jonallured: I guess there's one wrinkle here - Bearden stores all this stuff about an Organization, but when it comes time to export to RedShift, for example, we resolve that all down into a flattened CSV row. Bearden's purpose is to collage together bits of info and given some info about the rank of this info, return our best info about that Organization.

So, this leads me to think that for the API, what we should be sending back is this best guess - yes it's true that Bearden's schema is more complicated than what you see here, but what the clients should get back, IMHO, is the flattened org, not all the details.

alloy:
> I had one question about how we structure the API going forward - how nouns and verbs co-exist in the API. It doesn't seem intuitive to me to just put everything in a flat namespace.

Yes, where possible, don’t create root fields for every type or variation on that type, that’s gonna make the schema harder to explore.

We just discussed something like this yesterday [here](https://github.com/artsy/metaphysics/pull/770#issuecomment-334275108) where different feeds of artists are all made available through the single `artists` root field.

----

In case of the need for a root field to be able to retrieve an arbitrary node in the graph, add a root `node` field that, when given a globally unique ID, can return any type that implements the ‘Node’ interface https://facebook.github.io/relay/graphql/objectidentification.htm.

PS: I have some ideas about what these IDs should encode, with an eye on the future where we’ll stitch these schemas together, namely a prefix that will allow the proxy to identify the backend service it should delegate that request to `base64("bearden:organization:42")`.

spec/requests/api/v1/search_spec.rb
alloy: Are you asking for conventions around pagination? If so, this is what FB and GitHub use https://facebook.github.io/relay/graphql/connections.htm and we started using in MP later on. It’s all about using cursors, basically, and adds a convention for how to add metadata around a collection.

app/graphql/types/organization_type.rb
jonallured: Oh, looking at this again, now I guess what I'm proposing is more like this:

```ruby
Types::OrganizationType = GraphQL::ObjectType.define do
  name 'Organization'

  field :id, !types.ID
  field :name, !types.String
  field :tag_names, types[types.String]
  field :website_url, types.String
  field :city, types.String
  field :country, types.String
  field :in_business, types.Boolean
end
```

But in reality, the only thing we need to return at this point is name:

```ruby
Types::OrganizationType = GraphQL::ObjectType.define do
  name 'Organization'

  field :id, !types.ID
  field :name, !types.String
end
```

Because our only consumer at this point is an autocomplete field in Volt.

spec/requests/api/v1/search_spec.rb
jonallured: Yeah, that's good, but what I did for the API that this is replacing was support a param that decided how many records to return. The autocomplete won't need more than, say, 20 records. Do you have a way you think that should work? I guess you could use the cursor stuff, but is there an easier way?

app/graphql/types/organization_type.rb
alloy: Oh, in that case, I suggest returning just what you need. The best case scenario is that you only ever _add_ to the schema, so maybe don’t rush it when you don’t need certain fields yet?

app/graphql/types/query_type.rb
craigspaeth: I'm not caught up on the latest thinking in schema design, but the nice thing about having a schema that can evolve in an "append only" manner with auto-docs and not risking bloating response size, is you have more leeway to mix "RPC" and "resourceful" styles.

One way to think about designing GraphQL is in a resourceful manner which you can carry over your REST/CRUD design principles—so what would be POST/GET/PUT/DELETE would be...

```
query organization(id: 123) {}
query organizations(term: "david") {}
mutation createOrganization(name: "foo") {}
mutation updateOrganization(id: 123 name: "foo") {}
mutation deleteOrganization(id: 123 name: "foo") {}
```

In this case for search what would be `http://api.com/api/organizations?term="david"` in REST would be `organizations(term: "david")` in GraphQL.

That said, any style is fair game as GraphQL doesn't prescribe a style itself, but the community is [starting](https://blog.graph.cool/designing-powerful-apis-with-graphql-query-parameters-8c44a04658a9) to [talk](https://dev-blog.apollodata.com/designing-graphql-mutations-e09de826ed970) [about](https://www.youtube.com/watch?v=R8Ny0i5S978) [best practices](https://medium.com/impraise-design-engineering/making-mistakes-with-graphql-874b8ca62b9d).

spec/requests/api/v1/search_spec.rb
orta: That's fine, [this query](http://metaphysics-staging.artsy.net/?query=%7B%0A%20%20artist(id%3A%22banksy%22)%20%7B%0A%20%20%20%20artworks_connection(first%3A5)%20%7B%0A%20%20%20%20%20%20edges%7B%0A%20%20%20%20%20%20%20%20node%20%7B%0A%20%20%20%20%20%20%20%20%20%20title%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D&operationName=null) would return the first 5 elements on the connection 

```graphql
{
  artist(id:"banksy") {
    artworks_connection(first:5) {
      edges{
        node {
          title
        }
      }
    }
  }
}
```

Sidenote: We don't do this, but the FB reference GraphQL implementation [has both](http://graphql.org/swapi-graphql/?query=%7B%0A%20%20allFilms%20%7B%0A%20%20%20%20edges%20%7B%0A%20%20%20%20%20%20node%20%7B%0A%20%20%20%20%20%20%20%20title%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%20%20films%20%7B%0A%20%20%20%20%20%20title%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D&operationName=null) `{ edges: { node } }` for relay/apollo and a named version for convenience.

jonallured: Oh perfect, thanks! So then @dblock, what do you think about adding a spec for this?

app/graphql/types/query_type.rb
dblock: @alloy So this returns an empty array when nothing is found. Is that the same thing in GraphQL context? Does this mean it still needs a `!`? Where is the doc for this, I couldn't find it.

app/graphql/types/organization_type.rb
dblock: An organization can have multiple names. For now I'll return just `id` and `names`, but I think it's good to design complete APIs, otherwise we're designing for a specific client and constantly have to update both client and server to return new things without ever thinking about "what should this API return generically" vs. "what does this client need".

.rspec
dblock: Cause I like it better cause I can see what specs are failing better with debug output and such. But removing it if you feel strongly.

dblock:
@jonallured Regarding refactoring `GraphqlController` I think this isn't necessary because that query object is already `BeardenSchema`, there's literally no logic in `execute` other than parameter parsing from string to JSON which is really a request/response concern and should live in the controller.

.rspec
orta: We should build jest-rspec - then you get all the Jest tooling for Ruby

app/graphql/types/organization_type.rb
dblock: ... or do we think we should add `name` to the organization API and then pick the first? How is that going to work for a mutation?

spec/requests/api/v1/search_spec.rb
dblock: I added support for `first` with a spec.

app/graphql/types/query_type.rb
dblock: I am with @craigspaeth on the resourceful way of designing this. I'd like to do this in a subsequent PR and kill the search API. Again my goal for *this* PR was not to redesign the API but change the layout to be GraphQL.

alloy: Yeah, then you need to add it. There’s no specific documentation for this regarding collections (afaik), the important part is that in the schema `organizations` is just a field that returns something, it could be null or non-null like any other fields.

dblock:
@jonallured Again, all yours.

.rubocop_todo.yml
dblock: Yeah, this project is doing too much work for Rubocop. Usually we create a .rubocop.yml with `inherit_from: .rubocop_todo.yml` and run `rubocop -a ; rubocop --auto-gen-config` and do zero thinking about what should go in the main yml to avoid hurting our heads. I'll PR that separately.

app/graphql/types/query_type.rb
alloy: Not sure if you were only addressing @cavvia’s initial comment, but as far as making this a good example goes, you should really add the `!` back imo. The schema is now harder to use because it’s telling the clients that they need to perform a null check before using the results, which isn’t necessary.

app/graphql/types/organization_type.rb
jonallured: Yeah, what I'm saying is that an `Organization` record can have many of all it's associated types of details. A concept I've wrapped up in the `Rankable` module:

https://github.com/artsy/bearden/blob/master/app/models/rankable.rb

`Rankable` records should be resolved down to the highest ranked item and only that name should be returned.

.rspec
jonallured: Yeah, during development I sometimes like documentation too, but this setting goes all over, including on CI, where I really dislike it. I have a shell thingy for this too:

https://github.com/jonallured/dotfiles/blob/master/files/.zshrc.local#L7

.rubocop_todo.yml
jonallured: I totally disagree, sorry! I think a `.rubocop_todo.yml` is an antipattern - why run RuboCop if you're just going to ignore what it says!! 😝

app/graphql/types/query_type.rb
dblock: Added.

alloy: 🙏🏽

cavvia: I think @craigspaeth's resourceful approach (`organizations(term: 'david')`) is better than what we have now and is probably a more conventional schema design. If a GraphQL query is about selecting fields on objects, rather than simply nesting agnostic 'nodes', it makes sense to use it in a more resourceful manner, rather than mixing an RPC style in there with verbs such as 'search'. Having said that, mutates essentially look RPC in style! So it's a mixed bag.

@dblock Sure we can follow up with schema improvements in further PRs, as there's no live client yet, so we have that luxury.

app/graphql/types/organization_type.rb
cavvia: None of the data coming back from ES is ranked, ES just indexes unsorted arrays. My usual approach is to index all the attributes needed by search clients for performance purposes - like that we avoid additional trips to the DB. So in this case I would index a `display_name` separately from the names array and have ES just return that in the response. However this might not scale so well to all ranked attributes. But the alternative would be to do DB lookups and some live ranking, which sounds expensive to me. 

I do think my original comment still stands, we may well want to return actual Location objects at some point - e.g. to expose geocoded point information, and that would present a schema issue, since countries and cities are flattened here and not even related to each other. The search schema is generally optimized for filtering or searching on attributes, like country and city names, and not with other data modeling concerns.

cavvia: But sure, let's just return what we need for now. Indexing a separate `display_name` in ES is pretty trivial but could be a follow up PR.

dblock:
@jonallured LMK if you need anything else for this to merge

jonallured:
@dblock I'm not totally clear if @cavvia is satisfied with this stuff:

https://github.com/artsy/bearden/pull/311/files/de0ce8baba132562f99530244e5cf6510b2071dc#diff-5ee38b7ec5da5112d7d9ae2e97fd953d

Also, I still think it's weird for the endpoint to return all names for an organization. I think it should look like this:

```ruby
Types::OrganizationType = GraphQL::ObjectType.define do
  name 'Organization'

  field :id, !types.ID
  field :name, !types.String
end
```

Otherwise the client is left to fend for themselves, but without the most important piece of the puzzle - the ranks for these names! I think Bearden should resolve this and maybe that means we only index the winning name, but in any case, the API should return one name for an organization.

dblock:
That's fair, but my goal is not to redesign the API in this PR. The current API returns a bunch of stuff.

```ruby
expect(response_json).to eq(
          [
            {
              'names' => organization.names,
              'tag_names' => organization.tag_names,
              'website_urls' => organization.website_urls,
              'cities' => organization.cities,
              'countries' => organization.countries,
              'id' => organization.id,
              'in_business' => organization.in_business
            }
          ]
        )
```

 I already caved in and removed some fields, can we leave that to a future task?

jonallured:
Yep, that's good with me - as long as we get a 👍 from @cavvia on Monday, then I'll merge and we can follow up with that later. Thanks again!!

cavvia:
@dblock @jonallured That's good with me :+1:. Resolving the schema issues is something Jon & I can follow up with, as long as we fix it before the first client starts consuming we're fine.

PR #315, opened 2017-10-09
Fix search endpoint

jonallured:
When I merged the GraphQL work to master it was deployed to staging, so I went to go smoke test it there, but was getting this:

```ruby
NameError (uninitialized constant #<Class:0x007fd422b59b10>::Search
```

So the `autoload_paths` wasn't working as expected. I tinkered with this for a bit, but didn't make much progress. Instead, I'm just moving this file into the models folder and removing the namespace.

Open to other ideas, but this should allow me to deploy this work to staging and production.


jonallured:
Doing a selfie merge on this one to see if it fixes staging.

ashkan18:
👍

cavvia:
Interesting that this eager loading works on dev but not in staging/production.

dblock:
Strange.

yuki24:
More context: https://blog.bigbinary.com/2016/08/29/rails-5-disables-autoloading-after-booting-the-app-in-production.html

PR #317, opened 2017-10-13
@jonallured: fix graphql organization schema to use top ranked name #migration

cavvia:
Our representation of Organization objects via the Search API currently returns an unsorted array of names - API clients should only really see the top ranked name as this is the display name they should use when referring to the organization. The clients currently have no need for the alternate names, though we can add them to the schema if need be in future.

Here I index a `name` field in Elasticsearch as the authoritative (top ranked) name for an Organization, and additionally index other names in an `alternate_names` array as is our convention in other search indices. I've adjusted the search query to search on _all_ the names, but now the API only returns the `name` attribute. 

**Migration**

We need to recreate the index as we are scrapping the `names` field and adding the `name` and `alternate_names` fields in its pace. ES doesn't support deleting existing fields in a live index so in a console we drop the old index and re-declare the mappings for the new one:

```
Organization.reload_index!
```

And then we can dispatch our indexing sidekiq workers to populate the new index:

```
```

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -20,7 +20,8 @@ class Organization < ApplicationRecord
 
   searchable do
     field :id, type: 'integer'
-    field :name, using: :names, type: 'string', analysis: FULLTEXT_ANALYSIS, factor: 1.0
+    field :name, type: 'string', analysis: FULLTEXT_ANALYSIS, factor: 1.5
+    field :alternate_names, type: 'string', analysis: FULLTEXT_ANALYSIS, factor: 1.0
     field :tag, using: :tag_names, type: 'string', analysis: FULLTEXT_ANALYSIS, factor: 0.5
     field :website, using: :website_urls, type: 'string', analysis: FULLTEXT_ANALYSIS, factor: 0.5
     field :city, using: :cities, type: 'string', analysis: FULLTEXT_ANALYSIS, factor: 0.5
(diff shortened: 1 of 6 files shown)


app/models/organization.rb
```diff
@@ -34,8 +35,16 @@ def self.estella_search_query
...
+    names.first
+  end
+
+  def alternate_names
+    names.drop(1)
+  end
+
   def names
-    organization_names.pluck(:content)
+    organization_names.order(created_at: :desc).sort_by(&:rank).pluck(:content)
```
dblock: The `order` doesn't do anything given a `sort_by` here, don't you want to order by both in SQL?

dblock: I think all the other `_names` could benefit from a predictable sort as well, but that's beyond the scope of this PR.

jonallured: This looks correct to me - rank isn't stored in the database, it's computed on the fly and the created_at order breaks ties.

cavvia: Right, I would expect it to break ties and my spec confirms this - both the names there have the same rank.

jonallured:
This looks good to me - merging now and then I'll get it deployed and migrated out to production after I do a bundle update PR. 👍

app/models/organization.rb
yuki24: Keep in mind that Ruby's sort algorithm is not a stable sort (quick sort) and there could be an edge case that is extremely hard to find.

dblock: My spec disagrees. This doesn't break any ties, it re-sorts a collection previously sorted by `-created_at` by `rank`. See https://github.com/artsy/bearden/pull/321.

PR #320, opened 2017-10-16
@jonallured: ensure elasticsearch is up before running tests

cavvia:
There's currently no guarantee the Elasticsearch process is running when we start the test suite, leading to occasional build failures. It can take a while to initialize the ES container fully. This command will block till ES is available on localhost at the default port, which I believe is our test setup.


jonallured:
Awesome, thanks for the PR and the explanation!! ❤️

PR #321, opened 2017-10-17
Sort rankables by lowest rank first, then newest first, more explicitly.

dblock:
See https://github.com/artsy/bearden/pull/317#discussion_r144834710, just like @yuki24 said, sorting in the database, then outside is ignoring sorting in the database.

Consolidating ranking into `OrganizationResolver`.

What we want is to sort by rank (lowest rank first), then by time (newest first).

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -43,8 +43,9 @@ class Organization < ApplicationRecord
     names.drop(1)
   end
 
+  # Lowest rank first, then most recent first.
   def names
-    organization_names.order(created_at: :desc).sort_by(&:rank).pluck(:content)
+    OrganizationResolver.rank(organization_names).pluck(:content)
   end
 
   def tag_names
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,3 +37,6 @@ Rails/ReversibleMigration:
 
 Metrics/LineLength:
   Enabled: false
+
+Rails/TimeZone:
+  Enabled: false
--- a/app/models/organization_resolver.rb
+++ b/app/models/organization_resolver.rb
@@ -14,19 +14,23 @@ class OrganizationResolver
     highest_ranked_data
   end
 
+  def self.rank(rankables)
+    rankables.sort_by { |rankable| [rankable.rank, -rankable.created_at.to_i] }
+  end
+
+  def self.find_highest_rankable(rankables)
+    rank(rankables).first
+  end
+
   private
 
   def find_highest_rankables
-    @email = find_highest_rankable(@organization.emails)
-    @location = find_highest_rankable(@organization.locations)
-    @organization_name = find_highest_rankable(@organization.organization_names)
-    @organization_type = find_highest_rankable(@organization.organization_types)
-    @phone_number = find_highest_rankable(@organization.phone_numbers)
-    @website = find_highest_rankable(@organization.websites)
-  end
-
-  def find_highest_rankable(rankables)
-    rankables.order(created_at: :desc).sort_by(&:rank).first
+    @email = OrganizationResolver.find_highest_rankable(@organization.emails)
+    @location = OrganizationResolver.find_highest_rankable(@organization.locations)
+    @organization_name = OrganizationResolver.find_highest_rankable(@organization.organization_names)
+    @organization_type = OrganizationResolver.find_highest_rankable(@organization.organization_types)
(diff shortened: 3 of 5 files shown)


jonallured:
Yeah, the rules here are lowest rank is best and then rank ties go to the newest data. Consider the case where a gallery changes name. If we have a source that changes the name it reports, then we'd have data like this:

```
{ name: 'Busted Gallery', rank: 1, created_at: 1.year.ago }
{ name: 'Sparkle Gallery', rank: 1, created_at: 1.month.ago }
{ name: 'Dusted Gallery', rank: 2, created_at: 1.week.ago }
```

What we want is for `Sparkle Gallery` to be returned because it has the lowest rank and is the newest data. At this point I don't really care how the remaining records are sorted, but I would assume an order like this:

```
[
  { name: 'Sparkle Gallery', rank: 1, created_at: 1.month.ago },
  { name: 'Busted Gallery', rank: 1, created_at: 1.year.ago },
  { name: 'Dusted Gallery', rank: 2, created_at: 1.week.ago }
]
```

Sort first by `rank`, lowest first; then by `created_at`, newest first.

app/models/organization.rb
```diff
@@ -44,7 +44,7 @@ def alternate_names
   end
 
   def names
-    organization_names.order(created_at: :desc).sort_by(&:rank).pluck(:content)
+    organization_names.sort_by { |n| [-n.created_at.to_i, n.rank] }.pluck(:content)
```
cavvia: this query also exists [here](https://github.com/cavvia/bearden/blob/master/app/models/organization_resolver.rb#L29), so we would need to update that too.

spec/models/organization_spec.rb
```diff
@@ -91,6 +91,33 @@
...
+
+      PaperTrail.track_changes_with(raw_input_2) do
+        @n2 = Fabricate(:organization_name, organization: organization, created_at: Time.at(1))
+        @n3 = Fabricate(:organization_name, organization: organization, created_at: Time.at(2))
+      end
+
+      # n3 is created most recently
+      # n2 has a rank higher than n1, so it's last
+
+      expect(organization.names).to eq [@n3, @n1, @n2].map(&:content)
```
cavvia: as @jonallured describes, we would expect these sorted by `rank` first (ascending), `created_at` second. `n2` and `n3` come from a rank 2 source, with `n3` the newest, and `n1` comes from a rank 1 source. so the expected order is `[@n1, @n3, @n2]`, which is what I get when running this spec on master.

jonallured: Yeah, I think this might be a case of miscommunication about the requirements - we good to close this @dblock?

dblock: I don't understand why this works, `organization_names.order(created_at: :desc).sort_by(&:rank)` isn't "sort by date in descending order, then re-sort by rank", it should be a sort by rank after some pre-sorted order by created_at, which can potentially re-shuffle everything, but somehow doesn't?

yuki24: Regardless of the requirements here [the result is not guaranteed to be stable](https://stackoverflow.com/questions/15442298/is-sort-in-ruby-stable). It just happened to be stable. Stable sort [was once requested](https://bugs.ruby-lang.org/issues/1089) to Ruby core a long time ago but was rejected for performance reasons.

jonallured: Ok, I see the issue now - not sure how to write a failing spec for it, but as it stands, it's possible for sort instability to produce incorrect tie breakers. I think you had it right here @dblock:

```ruby
organization_names.sort_by { |n| [n.rank, -n.created_at] }.pluck(:content)
```

That looks like what we want so that instability can't mess us up. Do you want to update this PR with that approach?

Just had a thought - what about this:

```ruby
organization_names.order(created_at: :desc).each_with_index.sort_by { |name, index| [name.rank, index] }.pluck(:content)
```

It's a lot more verbose, but helps me see that the original sort should be the tie breaker. 🤷‍♂️

jonallured: One more thought: regardless of how we fix this, it's probably worth a comment reminding Future Jon about sort instability so he doesn't refactor this back to the buggy version. 😝

app/models/organization.rb
dblock: I consolidated ranking into `OrganizationResolver`.

dblock:
@jonallured This is updated and ready to merge.

jonallured:
Awesome, thank yous all around!! 🤗

PR #322, opened 2017-10-19
@jonallured: only allow jwt tokens from trusted apps

cavvia:
It turns out gravity has an [Auth API endpoint](https://github.com/artsy/gravity/blob/master/app/api/v1/me_auth_endpoint.rb#L20) that deals out JWT tokens to untrusted apps (like force) with a short expiry. This can theoretically be used to produce valid JWT tokens for Bearden knowing only the Bearden application ID, which is a security problem. 

While the Gravity API endpoint looks like a security hole, it can be useful for some apps who wish to allow access to a set of users from an untrusted app (like Impulse, which wants to allow some Force users access to their conversations inbox data, as Force is an untrusted app in this scheme). 

We've decided it's up to the destination app to validate app roles in its auth if it wishes to restrict access to trusted apps. This is the case with Bearden, so I've added the additional checks in our auth method to ensure that the app which requested the token is trusted. We use 'roles' embedded in the JWT to determine whether an app is trusted (see [gravity code](https://github.com/artsy/gravity/blob/master/app/models/util/application_trust.rb#L14)). 

cc @joeyAghion @mzikherman

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -10,11 +10,15 @@ class ApiController < ApplicationController
 
   def token_match?
     auth_header = request.headers['HTTP_AUTHORIZATION']
-    return false unless auth_header
     token = auth_header.split(' ').last
     secret = Rails.application.secrets.artsy_internal_secret
-    JWT.decode(token, secret)
-  rescue JWT::DecodeError
+    payload = JWT.decode(token, secret).first
+    validate_payload(payload)
+  rescue JWT::DecodeError, NoMethodError
     return false
+  end
+
+  def validate_payload(payload)
+    payload['roles'].split(',').include?('trusted')
   end
 end
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -10,15 +10,16 @@ class ApiController < ApplicationController
 
   def token_match?
     auth_header = request.headers['HTTP_AUTHORIZATION']
+    return false unless auth_header
     token = auth_header.split(' ').last
     secret = Rails.application.secrets.artsy_internal_secret
     payload = JWT.decode(token, secret).first
     validate_payload(payload)
-  rescue JWT::DecodeError, NoMethodError
+  rescue JWT::DecodeError
     return false
   end
 
   def validate_payload(payload)
-    payload['roles'].split(',').include?('trusted')
+    payload['roles']&.split(',')&.include?('trusted')
   end
 end
--- a/spec/graphql/search_spec.rb
+++ b/spec/graphql/search_spec.rb
@@ -3,7 +3,7 @@ require 'rails_helper'
 describe GraphqlController, type: :controller do
   let(:jwt_token) do
     JWT.encode(
-      { aud: Rails.application.secrets.artsy_application_id },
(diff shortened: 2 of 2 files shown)


dblock:
This highlights the need for a shared middleware across Artsy for these! cc: @orta 

I *think* the Gravity API is OK, it hands a token for a given app that can then be used to identify the user from gravity. That's authentication. Authorization is definitely the burden of this app and it should reject users it doesn't want.

I did raise this in #security on Slack btw. And please note that bearden is public, so this discussion is public as well.

joeyAghion:
So far apps have simply used `jwt` rather than shared middleware. Maybe it can be an extension of https://github.com/artsy/artsy-auth, which is session-oriented and has hooks for asserting user roles explicitly, but could be extended to encapsulate the JWT validation and expose the embedded app and role data.

app/controllers/api_controller.rb
```diff
@@ -10,11 +10,15 @@ def ensure_jwt
...
   def token_match?
     auth_header = request.headers['HTTP_AUTHORIZATION']
-    return false unless auth_header
     token = auth_header.split(' ').last
     secret = Rails.application.secrets.artsy_internal_secret
-    JWT.decode(token, secret)
-  rescue JWT::DecodeError
+    payload = JWT.decode(token, secret).first
+    validate_payload(payload)
+  rescue JWT::DecodeError, NoMethodError
```
yuki24: `NoMethodError` seems too generic here.

jonallured: I agree - any reason you didn't like the return false unless line @cavvia?

spec/graphql/search_spec.rb
```diff
@@ -3,7 +3,7 @@
 describe GraphqlController, type: :controller do
   let(:jwt_token) do
     JWT.encode(
-      { aud: Rails.application.secrets.artsy_application_id },
+      { aud: Rails.application.secrets.artsy_application_id, roles: 'trusted' },
```
jonallured: I'd like to see another test case here without the trusted role that shows it doesn't work - are you good with adding that?

app/controllers/api_controller.rb
cavvia: I agree - Rubocop was giving me an [AbcSize](www.rubydoc.info/gems/rubocop/0.27.0/RuboCop/Cop/Metrics/AbcSize) error, though I may have fixed that with the `validate_payload` extraction too. So can re-introduce and see if that works.

spec/graphql/search_spec.rb
cavvia: Done

app/controllers/api_controller.rb
cavvia: This is back now.

jonallured:
Thank you @cavvia!! 🥇

cavvia:
@dblock @joeyAghion Agreed a shared middleware would be good. It would need to support these trusted app tokens. I think we can probably get it into `artsy-auth`, even if that is more focused on  client devs.

jonallured:
FYI, I opened an issue over there artsy/artsy-auth#6 with some thoughts.

jonallured:
FYI, this is deployed to production.

PR #325, opened 2017-10-26
Lock to older redis version

jonallured:
While testing #324, I noticed some breakage and started investigating locally. Turns out there's a fixed issue with ActionCable and the new redis gem:

https://github.com/rails/rails/issues/30527

Until it's ironed out, I'm just going back to the older version.

There's also a small fix here for Bootstrap.


anandaroop:
Thanks for documenting this 👍

PR #328, opened 2017-10-27
Replace raw queries with Graphlient.

dblock:
This replaces raw queries with a parsed Graphlient query. This is much nicer because a spec can declare a complete query marked up with our favorite tooling (here we have `search` with two arguments, `term` and an optional `first`) and then execute it with all kinds of parameters without any strings involved. The result is a first class object of `OrganizationType`, so we can do `.name` for example. This also knows nothing about controllers and is identical to how a client would consume this API.

I separated things into a spec helper, an introspection spec and a search query spec which makes it easier to add future specs in the right place(s).

I think access denied with an invalid token is a 403 (Forbidden) as opposed to a 401 (token is valid, but access is denied). At least that's how Constellation does it, lets make it consistent one way or another.


jonallured:
LGTM! 💯

PR #334, opened 2017-12-04
Fix omniauth-oauth2 vulnerability alert

jonallured:
Unless I'm missing something, my only option here is to lock to a newer `omniauth-oauth2` version (1.4.0 is the latest) and allow `jwt` to be downgraded. Untangling these dependencies can be tricky, but from what I can tell, the problem with using a newer `jwt` is actually intridea/oauth2:

https://github.com/intridea/oauth2/blob/v1.4.0/oauth2.gemspec#L9

I found there are a few PRs trying to address some dependency issues there:

https://github.com/intridea/oauth2/pull/326
https://github.com/intridea/oauth2/pull/318
https://github.com/intridea/oauth2/pull/317

Skimming those issues makes it appear that Intridea isn't putting dev resources into the project anymore. The last commit was about 6 months ago. Open source, amiright?

Anyway, open to other ideas, but this seems like the right way to address the security warning.


ashkan18:
👍  looks good to me, and yes we may want to fix it in `artsy-auth`.

starsirius:
Context: The [security issue](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2012-6134) was alerted by Github [automatically](https://github.com/blog/2470-introducing-security-alerts-on-github).

screen shot 2017-12-04 at 12 18 21 pm

jonallured:
I'm going to go ahead and merge this - feel free to comment if you've got something to add!

PR #340, opened 2018-01-24
Use Heroku release phase for database migrations

jonallured:
I hadn't heard of this, but I like this approach much better for ensuring database migrations happen before the dynos are in production.

https://devcenter.heroku.com/articles/release-phase


yuki24:
Interesting!

PR #347, opened 2018-03-29
Upgrade Ruby

jonallured:
Upgrading to the latest Ruby with all those juicy, juicy security updates!!

Note: I expect this first build on Circle to fail because I'm asking for 2.5.1-node, which I don't believe exists, but I wanted to check. I'll fall back to 2.5-node and try that next.


jonallured:
With 56c658d I've switched to our own Artsy Ruby image. This will allow us to upgrade Ruby at our discretion instead of relying on Circle upgrading their Ruby images.

PR #357, opened 2018-06-07
Bundle update

jonallured:
Not too much to see here, but I was able to un-pin RuboCop, so that's nice.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -315,6 +315,7 @@ GEM
       concurrent-ruby (~> 1.0)
     inflecto (0.0.2)
     ipaddress (0.8.3)
+    jaro_winkler (1.5.1)
     jbuilder (2.7.0)
       activesupport (>= 4.2.0)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -454,7 +457,8 @@ GEM
       rspec-mocks (~> 3.7.0)
       rspec-support (~> 3.7.0)
     rspec-support (3.7.1)
-    rubocop (0.54.0)
+    rubocop (0.57.1)
+      jaro_winkler (~> 1.5.1)
```
yuki24: 🤔I wonder why rubocop now depends on `jaro_winkler`. Maybe they added some sort of "Did you mean?" feature?

jonallured: lol, how did you even notice that!! 🥇 😝

yuki24: turns out Rubocop has been doing it for quite a long time https://github.com/rubocop-hq/rubocop/pull/1293

